PR #335, opened 2015-07-20
added opencaching.ro to site list

following5:
workflow test / pull request test


wrygiel:
BTW - you should be able to commit directly. No need to make pull requests.

Issue #337, opened 2015-07-28
Update user.last_login on OKAPI login


following5:
The field `user.last_login` tracks the last login date on all OC installations. This field e.g. is important for rating the activity of cache owners when processing cache reports.

We currently have 2500 users in the OC.de database who have log entries that were created after the `user.last_login` date. ~2400 of them appeared after c:geo with OKAPI logging was made available, so it is confirmed that this issue is OKAPI related. (The other ~100 were before mid-2011, probably at this date some OC.de `last_login` related bug was fixed.)

To fix this, OKAPI should update `user.last_login` on successful Level 3 authentication.

wrygiel:
It all depends what "last login" means. For example, I have last logged in to my opencaching account a couple of months ago. I have simple never logged out since. So, is it really incorrect?

wrygiel:
Perhaps the field should be renamed to something completely else. Like "last_activity". However, this would mean that it would have to be updated upon almost every action.

Perhaps it should simply be removed instead? Seems a better alternative ;)

wrygiel:
BTW, you can easily check if a log entry was created by OKAPI by looking in the okapi_cache_logs table. I wouldn't be surprised if some of these logs were *not* created by OKAPI. In OCPL, there's a checkbox to "never log out" somewhere in user preferences.

following5:
The `user.last_login` field _has_ the meaning of "last activity", in OC.de and OC.pl code. It is updated if the user either explicitly logs in, or if the the login state is restored via cookie-stored session data, which happens on each HTTP access of any page which checks the login state. This is true for most pages of the OC site.

This field is important for assessing the mantainance state of cache listings. It's not a perfect way to to this, but an important indicator for this and frequently used by both OC admins and users/loggers.

> BTW, you can easily check if a log entry was created by OKAPI by looking in the okapi_cache_logs table

Done. It's 100% OKAPI.

I have no idea where to insert the "update user set last_login=now() where user_id=xxx" in the OKAPI oauth verification code. Can you do this, or give me an advice what's the proper place? It's critical for OC.de to get this fixed.

wrygiel:
Haha. :) Okay, I'm convinced.

following5:
oops

wrygiel:
I think you may put this in the OkapiServiceRunner::call method. If $request contains token, then it is an authenticatad user request. You can then update the proper field in the database.

[...]

PR #338, opened 2015-07-28
update user.last_login when running level 3 methods; updates #337

following5:
I have tested this with `>= 3` changed to `>= 1` and `$request->token->user_id` replaced by a constant number, as I currently do not have any oauth enabled OKAPI client available for testing.


wrygiel:
Seems correct! Merge it. :)

BTW: OKAPI->OCPL synchronization is temporarily disabled so you won't see it in action on OCPL. Commits to OKAPI repo will commit to old OCPL SVN repo, but the live server is in now sync with the new GitHub repo.

BTW2. You can try this tool for testing OAuth - https://github.com/wrygiel/okapi-browser

PR #359, opened 2016-01-15
restored xmlns:ox definition; fixes #347

### LINKED ISSUE #347: opencaching.com is gone - update ns_ox
The GPX files created by OKAPI refer to the now nonexistent http://www.opencaching.com/xmlschemas/opencaching/1/0/opencaching.xsd, as well as the docs on services/caches/formatters/gpx which have an additional link to http://www.opencaching.com. I assume that the nx_ox extension should not be dropped for compatibity reasons.

The web archive has the XSD: http://web.archive.org/web/20150512043245/http://www.opencaching.com/xmlschemas/opencaching/1/0/opencaching.xsd

following5:
please review

### DIFF
--- a/okapi/services/caches/formatters/gpxfile.tpl.php
+++ b/okapi/services/caches/formatters/gpxfile.tpl.php
@@ -11,7 +11,7 @@ echo '<?xml version="1.0" encoding="utf-8"?>'."\n";
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="
 http://www.topografix.com/GPX/1/0 http://www.topografix.com/GPX/1/0/gpx.xsd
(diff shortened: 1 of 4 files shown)


okapi/services/caches/formatters/gpxfile.tpl.php
```diff
@@ -11,7 +11,7 @@
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="
 http://www.topografix.com/GPX/1/0 http://www.topografix.com/GPX/1/0/gpx.xsd
-http://www.opencaching.com/xmlschemas/opencaching/1/0 http://www.opencaching.com/xmlschemas/opencaching/1/0/opencaching.xsd
+http://www.opencaching.com/xmlschemas/opencaching/1/0 https://github.com/opencaching/okapi/blob/master/etc/nsox.xsd
```
wrygiel: I think that you need a raw URL here. The one you have here will output HTML (not XML Schema), so the validators won't recognize it.

following5:
ok

PR #365, opened 2016-01-28
optimized search_nearest; closes #363

### LINKED ISSUE #363: Optmize search_nearest
The OKAPI method search_nearest produces most of the server load of Opencaching.de. It is rather slow (0,2 seconds per call) because the condition

`$where_conds[] = "$distance_formula <= '".mysql_real_escape_string($radius)."'";`

does not allow any optimization by the SQL server. _All_ caches which meet the other conditions need to be searched. But there is a latitude/longitude index on the ´caches` table which should allow to narrow down the search set for the typical small radiuses.

What about some bounding box condition which is evaluated first? For an 1 km radius in Germany, a simple longitude filter would narrow down the number of caches by ~ factor 700. Only 100 caches need to be checked for the radius and any other search filters. I think it should be worth to apply such a longitude prefilter up to search radiuses of 100 km.

(Btw, the OKAPI stats records are really nice for such performance evaluations :-)

following5:


wrygiel:
Thanks, I will try review it tomorrow!

PR #378, opened 2016-03-08
added full support for OCDE 'needs maintenance' flag/options

following5:
The feature is online meanwhile on OC.de, and I think c:geo developers are ready to implement it - together with submitting NM logs in OCPL sites.

### DIFF
--- a/okapi/services/caches/formatters/gpx.php
+++ b/okapi/services/caches/formatters/gpx.php
@@ -211,7 +211,7 @@ class WebService
         if ($vars['images'] != 'none')
             $fields .= "|images";
         if (count($vars['attrs']) > 0)
-            $fields .= "|attrnames|attr_acodes";
+            $fields .= "|attrnames|attr_acodes|needs_maintenance";
         if ($vars['trackables'] == 'desc:list')
             $fields .= "|trackables";
         elseif ($vars['trackables'] == 'desc:count')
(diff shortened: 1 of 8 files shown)


wrygiel:
I am not sure about the meaning of the `needs_maintenance` field in `logs/entry`. I'd like to make it a little more explicit in meaning. I'll upload an example soon.

wrygiel:
One question (just confirming if I got this right):

What happens if 10 users say that a cache needs maintenance, and then one user says it doesn't? Is the flag cleared? Is it remotely possible that in future this will evolve into something more sophisticated, e.g. a voting system?

following5:
The maintenance state of a cache is momentarily, it can change one minute later. So only the flag of the last log - ordered by log date/time - is relevant.

wrygiel:
So, the flag can be set multiple times, but can be unset only once? What would happen if two users attempted to unset the flag concurrently?

wrygiel:
Or perhaps it is possible to unset the flag multiple times too?

wrygiel:
If an "unsetting" log entry gets deleted, is the flag set again?

following5:
Any log entry can contain any flag, set or reset. So there can be multiple NM-flag logs in a row or multiple no-NM-flags in a row. These logs just say: A the _time of this log_ the cache hat _that_ state.

| If an "unsetting" log entry gets deleted, is the flag set again?

Yes, if the newest of the remaining log entries with NM != false is NM = true. The cache's NM flag always reflects the flag of the last visible log with NM != false, given the log order `date` desc, `date_creeated` desc, `id` desc. This field ist just an optimization, so that the log table needs not to be scanned on each request.

wrygiel:
The distinction between `false` and `clear` is not implicit enough (users will need to study documentation in order to get it). Perhaps we can do this a little better. I will submit my proposal soon.

wrygiel:
Posted it here: https://github.com/following5/okapi/pull/1

following5:
* added needs_maintenance2 to logs/submit
* changed logs/entries field to strange past-tense logic - "needed_maintenance"
* clarified tensens in documentation of logs/entry

I have named the NM options "yes", "no" and "null" because this avoids errors when migrating from logs/subit `needs_maintenance`. Ok?

This change is not tested yet.

wrygiel:
> changed logs/entries field to strange past-tense logic

That's true - I also find this commit very strange. I think that we have completely misunderstood each other. The way you propose it now seems to me worse than your original proposal. I will commit what I meant to a local branch - it will be easier to work together there.

wrygiel:
Exaggerated a bit. Some parts of your commit are very good. But the past tense in some other parts was strange indeed :)

following5:
Probably I did some [DOIP](https://en.wikipedia.org/wiki/WP:DWIP)

PR #380, opened 2016-03-10
Fix #373 - merge "feature/needs_maintenance2"

### LINKED ISSUE #373: Add OCDE "needs maintenance" flags/status
Opencaching.de will soon add the feature to submit the flags "geocache needs maintenance" and "geocache listing is outdated" when logging a cache. This will set the new geocache properties "needs maintenance since" and "listing is outdated since" to an appropriate date, or reset them to null.

Adding this properly to OKAPI will need lots of detail solutions and bells & whistles. (Maybe I can simplify it by moving some common code into database trigger functions). I think it is easiest if I do a sample implementation.

wrygiel:


wrygiel:
Hmmm.... Do you know how to name a pull request so that GitHub "knows" that it fixes #373? Trying to set a proper name, but none of these work :)

wrygiel:
BTW, I still haven't tested this. I think we may postpone testing until we find a middle ground for our naming problems. :)

wrygiel:
I'm off to sleep for today!

following5:
> Hmmm.... Do you know how to name a pull request so that GitHub "knows" that it fixes #373? Trying to set a proper name, but none of these work :)

The "fix #373" needs to be in a commit comment. It then will close isue 373 upon merge.

following5:
Except the the things annotated above, I am fine with this implementation.

following5:
With the commits above, I think everything is fine. Have tested it on OCDE and OCPL develsite. I won't insist on documenting the OCDE feature of confirming "Archived" and "Locked" cache states. :)

I think that submitting `needs_maintenance2 = false` may also be implemented for OCPL. It works like `needs_maintenance2 = true`, with two differences:

* It produces an "Maintenance performed" [second] log entry instead of "Needs maintenance".
* It is allowed only for the cache owner.

However I do not fully understand how this feature is used at OCPL sites, therefore I rather would not add it myself.

following5:
With the commits above, I think everything is fine. Have tested it on OCDE and OCPL develsite. I won't insist on documenting the OCDE feature of confirming "Archived" and "Locked" cache states. :)

I think that submitting `needs_maintenance2 = false` may also be implemented for OCPL. It then works like `needs_maintenance2 = true`, with two differences:

* It produces an "Maintenance performed" [second] log entry instead of "Needs maintenance".
* It is allowed only for the cache owner.

However I do not fully understand how this feature is used at OCPL sites, therefore I rather would not add it myself.

wrygiel:
> With the commits above, I think everything is fine. Have tested it on OCDE and OCPL develsite.

I'm fine with them too.

PR #385, opened 2016-03-14
adjusted log entry ordering for OCDE

following5:
The OKAPI docs are not specific about how the `latest_logs` are ordered. I think it won't break anything when they are not orderd by the `date` field. The new OCDE `order_date` field must not be disclosed for privacy reasons.

### DIFF
--- a/okapi/services/caches/geocaches.php
+++ b/okapi/services/caches/geocaches.php
@@ -703,17 +703,32 @@ class WebService
             foreach ($results as &$result_ref)
                 $result_ref['latest_logs'] = array();
 
-            # Get all log IDs with dates. Sort in groups. Filter out latest ones. This is the fastest
-            # technique I could think of...
+            # Get all log IDs in proper order, then filter out the latest
+            # ones. This should be the fastest technique ...
+
+            # OCDE allows to submit logs without time. To minimize problems
+            # when ordering logs with and without time on the same day, there
+            # is a separate order_date field which is caluclated from 'date'
+            # and 'date_created'.
+            #
+            # OCPL log entries are all submitted with time, so the 'date'
+            # field is sufficient for ordering.
+
+            if (Settings::get('OC_BRANCH') == 'oc.de') {
+                $logs_order_field_SQL = 'order_date';
+            } else {
+                $logs_order_field_SQL = 'date';
+            }
 
             $rs = Db::query("
-                select cache_id, uuid, date
+                select cache_id, uuid
                 from cache_logs
                 where
                     cache_id in ('".implode("','", array_map('\okapi\Db::escape_string', array_keys($cacheid2wptcode)))."')
                     and ".((Settings::get('OC_BRANCH') == 'oc.pl') ? "deleted = 0" : "true")."
-                order by cache_id, date desc, date_created desc
+                order by cache_id, ".$logs_order_field_SQL." desc, date_created desc, id desc
             ");
+
             $loguuids = array();
             $log2cache_map = array();
             if ($lpc !== null)
@@ -724,10 +739,6 @@ class WebService
                     $tmp[$row['cache_id']][] = $row;
                 foreach ($tmp as $cache_key => &$rowslist_ref)
                 {
-                    usort($rowslist_ref, function($rowa, $rowb) {
-                        # (reverse order by date)
-                        return ($rowa['date'] < $rowb['date']) ? 1 : (($rowa['date'] == $rowb['date']) ? 0 : -1);
-                    });
                     for ($i = 0; $i < min(count($rowslist_ref), $lpc); $i++)
                     {
                         $loguuids[] = $rowslist_ref[$i]['uuid'];
@@ -744,6 +755,8 @@ class WebService
                     $log2cache_map[$row['uuid']] = $cacheid2wptcode[$row['cache_id']];
                 }
             }
+
+            Db::free_result($rs);
 
             # We need to retrieve logs/entry for each of the $logids. We do this in groups
             # (there is a limit for log uuids passed to logs/entries method).


okapi/services/caches/geocaches.php
```diff
@@ -703,17 +703,32 @@ public static function call(OkapiRequest $request)
...
 
-            # Get all log IDs with dates. Sort in groups. Filter out latest ones. This is the fastest
-            # technique I could think of...
+            # Get all log IDs in proper order, then filter out the latest
+            # ones. This should be the fastest technique ...
+
+            # OCDE allows to submit logs without time. To minimize problems
+            # when ordering logs with and without time on the same day, there
+            # is a separate order_date field which is caluclated from 'date'
+            # and 'date_created'.
```
wrygiel: Just out of curiosity: Is it updated by a trigger?

okapi/services/caches/geocaches.php
```diff
@@ -703,17 +703,32 @@ public static function call(OkapiRequest $request)
...
 
             $rs = Db::query("
-                select cache_id, uuid, date
+                select cache_id, uuid
                 from cache_logs
                 where
                     cache_id in ('".implode("','", array_map('\okapi\Db::escape_string', array_keys($cacheid2wptcode)))."')
                     and ".((Settings::get('OC_BRANCH') == 'oc.pl') ? "deleted = 0" : "true")."
-                order by cache_id, date desc, date_created desc
+                order by cache_id, ".$logs_order_field_SQL." desc, date_created desc, id desc
```
wrygiel: The current version of the `order by` clause was designed to exactly match one of the indices. I'm not sure if adding ID won't cause MySQL to do something stupid (you probably know that their query optimizer is far from perfect). Is the ID really necessary?

okapi/services/caches/geocaches.php
following5: yes

okapi/services/caches/geocaches.php
wrygiel: It *probably* won't hurt, but still, we should at least take a look at the stats after a week or so.

following5: Uh, where are the line comments that I included in the pull request? There is neither a matching index for the old nor for the new order at OCPL, see http://opencaching.pl/okapi/devel/dbstruct. Besides of that, I think that an _additional_ order field will not affect the index selection; the server will use the index for the fields matching fromt he left, and then sort the result for the remaining fields.

Having that said: The ID currently is not necessary for OCPL, but for OCDE (because of the order of needs_maintenance2=true and needs_maintenance2=false logs).

following5:
Can you have a look at my comments above?

okapi/services/caches/geocaches.php
```diff
@@ -724,10 +739,6 @@ public static function call(OkapiRequest $request)
                     $tmp[$row['cache_id']][] = $row;
                 foreach ($tmp as $cache_key => &$rowslist_ref)
                 {
-                    usort($rowslist_ref, function($rowa, $rowb) {
-                        # (reverse order by date)
-                        return ($rowa['date'] < $rowb['date']) ? 1 : (($rowa['date'] == $rowb['date']) ? 0 : -1);
-                    });
```
wrygiel: Huh. Indeed. Seems completely redundant. Must have missed it during some refactoring...

okapi/services/caches/geocaches.php
wrygiel: > There is neither a matching index for the old nor for the new order at OCPL, see http://opencaching.pl/okapi/devel/dbstruct

This one (note, that the `deleted` field is always present in the `where` condition in OCPL, so it should match):

```
KEY `by_cache` (`cache_id`,`deleted`,`date`,`date_created`)
```

> Besides of that, I think that an additional order field will not affect the index selection

In general, you're right. But MySQL sometimes does stupid things. ;)

wrygiel: I think we can commit it as that, and check the call times afterwards.

wrygiel: Copied from 3dbe476:

> I verified that this ordering is index-optimized on OCDE database, but _no_ index is used at OCPL. This was also the case for the original order statement. there is a cache_id/deleted/date/date_created index at OCPL, but it will only be used if `FORCE INDEX (by_cache)` is added.

Oh, so you're **sure** the index is not used in OCPL?

following5: > This one (note, that the deleted field is always present in the where condition in OCPL, so it should match):

I tested that, and the `explain` said that no index is used at OCPL. (Hovever on MariaDB.)

following5: I mean: I tested it on MariaDB with current OCPL database structure.

wrygiel: <pre>
mysql> explain select cache_id, uuid from cache_logs where cache_id in (1,2,3,4,5) and deleted = 0 order by cache_id, date desc, date_created desc;
+----+-------------+------------+-------+---------------+----------+---------+------+------+-----------------------------+
| id | select_type | table      | type  | possible_keys | key      | key_len | ref  | rows | Extra                       |
+----+-------------+------------+-------+---------------+----------+---------+------+------+-----------------------------+
|  1 | SIMPLE      | cache_logs | range | by_cache      | by_cache | 6       | NULL |  540 | Using where; Using filesort |
+----+-------------+------------+-------+---------------+----------+---------+------+------+-----------------------------+
1 row in set (0.00 sec)
</pre>

wrygiel: Do you get anything better on OCDE?

following5: No, sorry. I must have made some mistake when testing on the OCPL database. :(

I have just verified that both MySQL (5.0) and MariaDB (5.5) are ok with the additional 'id' order in OCPL database.

PR #386, opened 2016-03-15
added listing_is_outdated field to logs/entries

following5:
Totally forgot about this. It is not necessary that "listing is outdated" is submitted through OKAPI, but it would be useful if this log property can be retrieved.

The code change does not need review, but you may want to verify the definition and documentation of the newly added field.

### DIFF
--- a/okapi/services/logs/entries.php
+++ b/okapi/services/logs/entries.php
@@ -25,6 +25,7 @@ class WebService
     private static $valid_field_names = array(
         'uuid', 'cache_code', 'date', 'user', 'type', 'was_recommended', 'comment',
         'images', 'internal_id', 'oc_team_entry', 'needs_maintenance2',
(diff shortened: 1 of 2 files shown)


wrygiel:
> The code change does not need review, but you may want to verify the definition and documentation of the newly added field.

Thanks for waiting! It seems okay!

wrygiel:
I guess you rebased and fast-forwarded? :) Why? You don't like merge-commits? ;)

wrygiel:
They are of use sometimes, i.e. they leave a reference to a pull request in git history. This particular pull request isn't very interesting, but some others are.

following5:
> They are of use sometimes, i.e. they leave a reference to a pull request in git history. 

Ok. I was not aware of that, because I usually operate git completely from the command line and therefore merge locally.

Issue #360, opened 2016-01-15
Submit log pictures


following5:
Submitting log pictures via OKAPI would be a great enhancement.

The are some differences in image postprocessing by OCPL and OCDE, with the OCDE processing afaik being more refined, e.g. not shrinking pictures which have a small file size, and resolving EXIF rotation if the Imagick PHP extension is available. I think it should be fine to adopt the OCDE implementation for all nodes.

Maximum image dimensions and file size threshold need to be configured, as well as picture upload path and maybe allowed file types.

wrygiel:
+1

How external developers will know about restrictions? Extra fields in apisrv/installation?

following5:
They need not to know about size limits for the stored images, because pictures will automatically be shrunk. The file formats accepted by OKAPI can be defined to be the same for all nodes (in the simplest case, just JPEG), so this also is no issue.

Another limit are the Megapixels of the uploaded original image, because they consume lots of memory while rotating/shrinking. This is difficult to handle, because the memory usage depends on the PHP graphics library implementation, which may be updated between any two PHP script executions. OC.de just puts a conservative limit on the file size; the shrink/rotate script may still crash e.g. for a 40 MP blue-sky picture. (Actually, within one and a half year this never happened.) It needs some thought and some tests to decide how to handle this in OKAPI.

wrygiel:
Yes, dimentions are not important, but size is. If we don't restrict it in any way, this will ultimately lead to developers sending us 30+ MB files (which will then be shrunk to 100 kB).

following5:
Ok. OCDE has a settings for this (currently 6 MB), OCPL a constant of 3.5 MB.

following5:
Btw, the pixel dimensions of an image can be retrieved without processing the whole file; it is in the header. As the temporarily uploaded file size doesn't matter and the real limit is PHP memory, a megapixel limit would be the most effective soution on the server side. Also for client apps which directly access the device's camera and can chose an MP limit upfront. But for anything else, the file size limit will be easier to handle ...

wrygiel:
It's also easier on the user's bandwidth.

How about adding a set of `rcmd_*` and `max_*` fields to the `apisrv/installation` method?

following5:
I don't know what rcmd_ is for. Adding the site properties to apisrv/installation should be fine.

What about these methods:

logs/add_image   to add an image to a log (title, image data, spoiler flag)
logs/edit image   to change the title and/or spoiler flag of an image
[...]

PR #392, opened 2016-03-18
Feature/logimages

following5:
This no pull request, but a discussion thread for implementing #360. Looks like this is the best way to collect all commit-related information for a feature branch at one place.


following5:
Are you sure that you want to encourage client developers to work around the OCPL restriction of forcing users to enter image captions? If yes, I think it should be documented that OKAPI may generate dummy captions if no caption is provided. Developers may rather want to force users to enter a caption than producing dummies.

If no, then there is no need to make things that complicated now. A required parameter still can be changed to 'optional' later, if necessary.

following5:
All those OKAPI featues that work different on different OC sites can discourage developers from implementing them at all. :( At least as long there is no api to query all those site capabilities and parameters.

wrygiel:
> Are you sure that you want to encourage client developers to work around the OCPL restriction of forcing users to enter image captions?

What do you mean?

> If yes, I think it should be documented that OKAPI may generate dummy captions if no caption is provided. Developers may rather want to force users to enter a caption than producing dummies.

For now, OKAPI won't generate dummy captions, and it will require a caption to be entered. But this interface gives us the ability to change that in various ways in the future.

> If no, then there is no need to make things that complicated now. A required parameter still can be changed to 'optional' later, if necessary.

But the existing applications won't respect that then (they will still require the caption provided by their user).

following5:
> Looks like this is the best way to collect all commit-related information for a feature branch at one place.

The cleaner way is to include "update #issue" in all commit comments, so they will have a backlink to the issue. But it's easy to forget that ...

wrygiel:
> The cleaner way is to include "update #issue" in all commit comments, so they will have a backlink to the issue. But it's easy to forget that ...

I think pull request is okay in this case. We have a separate branch we're working on, and we are going to merge it once we're done. That's exactly what pull requests are for :)

following5:
> Approximately 20% of these images have useful captions. The rest (80%) do not contain any useful information, just meaningless comments, or random strings.

There is little space for captions in the OCPL log image galleries - not enough for descriptive captions. So the loggers enter garbage.

following5:
AI have used the most common GD library for image processing. But it would be nice to use _Imagick_ instead if available - it would allow to strip EXIF data from JPEGs without resampling the images (which loses some quality).

following5:
With the last commit, I added a table okapi_images, which works like okapi_cache_logs. But now I have realized that there can be many other things submitted through OKAPI, e.g. the geocache reports. Instead of cluttering the database with history tables, it seems more efficient to put that all into one table.

So I will try to change that now. If it's not ok, it can just be reverted.

okapi/core.php
```diff
@@ -1902,6 +1904,26 @@ public static function fix_oc_html($html)
 
         return $html;
     }
+
+    function php_ini_get_bytes($setting)
+    {
+        $setting = ini_get($setting);
+        $setting = str_replace('K', '*1024', $setting);
+        $setting = str_replace('M', '*1024*1024', $setting);
+        $setting = eval('return '.$setting.';');
```
wrygiel: Are you sure it's safe to use eval in this context? Perhaps `preg_replace("/^[0-9*]+$/", $setting)`?

wrygiel: (btw, I still haven't read all of your emails, sorry - I decided I will first take a look at the diff)

okapi/settings.php
```diff
@@ -180,6 +185,21 @@
...
+        /**
+         * Maximum 'resolution' of saved images in pixels
+         */
+        'IMAGE_MAX_PIXEL_COUNT' => 524288,  # 0.5 MP
+
+        /**
+         * Quality of saved JPEG images on a scale from 50 to 100. The higher the
+         * quality, the sharper the image and larger the saved image files
+         * (file size grows exponentially). The PHP default is 75, which is also
+         * used in OCPL code. As OKAPI always resamples uploaded images - which
```
wrygiel: OKAPI always reasamples - why? :) Probably there's a thread for this, if so, I will find it in a minute.

wrygiel:
> Instead of cluttering the database with history tables, it seems more efficient to put that all into one table.

Ok. I'm fine with both approaches for now.

wrygiel:
I won't have time to review all the code now, but I had a look at all the changes to the interface.

The only thing I'm not entirely comfortable about is lack of the possibility to enforce max upload sizes. It's like saying "you can always push anything you want at us, regardless of its size, and we're required to always support this, indefinitely".

wrygiel:
If we keep it like that, then - in theory - we should support all POST sizes. The external programmer has no way of knowing what the limits of our server are.

following5:
> OKAPI always reasamples - why? :) Probably there's a thread for this, if so, I will find it in a minute.

You fill find it in several places, e.g. the comment above on GD and EXIF. To avoid resamling JPEG for images below pixel limit, one of the following would need to be done:

* Keep EXIF information. This e.g. can lead to accidential spoilering of final coordinates for multies and mysteries, because many modern cameras record GPS location; or to unexpected disclosure of private data,
* use Imagick instead of GD, which currently is only installed at OC.de,
* require some other library for stripping EXIF, or a binary tool where the data is piped through.

following5:
> If we keep it like that, then - in theory - we should support all POST sizes. The external programmer has no way of knowing what the limits of our server are.

There are limits on everything. E.g. the maximum size of a log comment, or the maximum amount of additional OKAPI requests the server can handle just now. And of course the maximum pixel dimensions of an image. But currently none of those limits are exposed.

But if you can implement such a limit in some meaningful way, I won't mind.

following5:
> The only thing I'm not entirely comfortable about is lack of the possibility to enforce max upload sizes. 

_How_ do you want to enforce such a limit, technically?

following5:
Note: All OC sites currently resample all uploaded images and strip EXIF data. OCPL resamples with quality 75, OCDE with quality 85 (see the JPEG_QUALITY setting in settings.php).

following5:
> If we keep it like that, then - in theory - we should support all POST sizes. The external programmer has no way of knowing what the limits of our server are.

The problem is - I don't know a way to know the limits either. I set PHP `post_max_size` to 1 MB, verified that the setting was reflected by `phpinfo()`, and still could send a complete 1.6 MB base64-encoded image to OKAPI.

Maybe you have more luck in implementing this. :)

okapi/core.php
following5: It is safe as long as noone with root rights enters something evil in php.ini. If someone does that, we probably have larger problems than this :)  but I like regex and have inserted one which verifies the setting. This also protects from requesting a non-numeric variable here.

The alternative to eval() would be extracting the number and (optional) multiplier and do the maths in direct PHP code.

wrygiel: Oh, I didn't notice that it's just from php.ini, sorry. I thought that you parse the value passed in `$setting`.

following5:
I am not sure if you noticed this: I did a change to the definition of logs/images/edit, requiring that at least one of the optional parameters is given. The main reason is that I did not know what `success` message to return in this case: "Your image has successfuly not been updated"?

The overall implementation is complete, and I think I have tested everything on both OCDE and OCPL code.

wrygiel:
> The main reason is that I did not know what success message to return in this case: "Your image has successfuly not been updated"?

The most probably scenario of this occurring would be that of a user clicking "edit" and then clicking "save" on some form without editing the form. I guess that in this case the message "successfully updated" is still okay.

I will have a last look and confirm there's nothing wrong with the interface.

If you want, I can also try to review the code, but this would take longer.

wrygiel:
Done. Made some descriptions more explicit. Check them out to make sure if I understand them the way you wanted them understood. If yes, then feel free to merge.

following5:
I have opened an [OCPL pull request](https://github.com/opencaching/opencaching-pl/pull/430) for updating okapi_settings.php.

PR #394, opened 2016-03-19
Improved example for order_by parameter

bartekchlebek:
In the existing documentation it may be unclear whether multicolumn sorting with two different orderings is supported. This example makes it explicit.


wrygiel:
Thanks!

PR #405, opened 2016-04-03
started changelog

following5:
This is a sample implementation for the new OKAPI changelog.

If you decrease the version number in core.php, the changelog will split up in a section with unavailable changes and a section with available changes on the site.

### DIFF
--- a/etc/changes.xml
+++ b/etc/changes.xml
@@ -8,6 +8,11 @@
 
     New entries are inserted at the top with EMPTY version number and date.
     Then run update_changes.php to insert the version numbers and dates.
+
+    Only one entry per commit. If multiple things were changed with one
+    commit, put them into one entry. This helps to detect copy-and-paste
+    errors, and allows commmit IDs or version numbers to be used as unique
+    identifiers for changelog entries.
    
     -->
     <changes>
@@ -71,9 +76,7 @@
             Improved posting plaintext logs on OCPL sites.
         </change>
         <change commit="28825396" version="1125" time="2015-12-22T19:12:16+0100" type="enhancement">
-            Enabled full 32-bit Unicode for submitting logs on some OC sites.
-            Use <a href="services/apisrv/installation.html">services/apisrv/installation</a>
-            to find out the <em>submit_charset</em> of an OC site.
+            Enabled full 21-bit Unicode for submitting logs on some OC sites.
         </change>
         <change commit="b2d31a34" version="1119" time="2015-11-26T14:11:51+0100" type="bugfix">
             Solved rare problem when searching by distance or bbox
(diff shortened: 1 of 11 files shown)


okapi/views/changelog.php
```diff
@@ -0,0 +1,52 @@
...
+use okapi\OkapiHttpResponse;
+use okapi\views\menu\OkapiMenu;
+
+class View
+{
+    public static function call()
+    {
+        require_once($GLOBALS['rootpath'].'okapi/views/menu.inc.php');
+
+        # TODO: load and cache from OKAPI repo
```
wrygiel: Where's the caching part? :)

etc/changes.xml
```diff
@@ -0,0 +1,67 @@
...
+or site admins.
+
+New entries are inserted at the top with EMPTY version number
+and date. Run update_changes.php to insert the version numbers
+and dates.
+
+-->
+<changes>
+    <change commit="93cfc2d9" version="1237" date="2016-03-31" type="enhancement">
+        New methods for &lt;a href="services/logs/images/add.html"&gt;adding&lt;/a&gt;,
```
wrygiel: Wouldn't it be easier if we allowed inline XHTML here (`<` instead of `&lt;`)? As we do in XML files for method descriptions?

okapi/views/changelog.php
```diff
@@ -0,0 +1,52 @@
...
+        # TODO: verify XML scheme
+        $changes = simplexml_load_file('https://raw.githubusercontent.com/following5/okapi/feature/changelog/etc/changes.xml');
+
+        $unavailable_changes = array();
+        $available_changes = array();
+
+        foreach ($changes->change as $change) {
+            if ($change['version'] > Okapi::$version_number)
```
wrygiel: I would also add a safeguard here: there's no version or date, skip the entry. (In case one forgets to run `update_changes.php`.)

etc/changes.xml
following5: Yes, I will change it.

okapi/views/changelog.php
following5: done

okapi/views/changelog.php
following5: An exception will be raised now if version or date is missing or a commit is duplicate (copy&paste error). Also, wrong versions and dates will be detected and replaced (copy&paste errors).

following5:
I am finished with the implementation, ready to merge.

following5:
Are there any issues left with the changelog that need further discussion?

wrygiel:
Probably not. I still haven't actually seen it (didn't work on my VM, but I didn't have time to debug why), but this part is not hindered by backward-compatibility issues in any way (we can change it later), so feel free to merge it.

following5:
> this part is not hindered by backward-compatibility issues in any way (we can change it later)

jep

PR #406, opened 2016-04-03
inform developers about the submit-charset

following5:


wrygiel:
Perhaps it would be easier to convert OCPL to the newer utf instead? If you can tell me what requires changing, I can forward it to the OCPL team (or maybe even do it myself).

following5:
I think the easiest solution would be to completely migrate OCPL to utf8mb4. This would mean:

* Update the MySQL settings to enable utf8mb4 charset on all OCPL sites,
* [migrate all tables on all sites to utf8mb4 charset and use "set names utf8mb4" instead of "... utf8"](https://github.com/OpencachingDeutschland/oc-server3/commit/fcd70054b82ecaecd3815b8dc3ddda68f54ac433),
* change the TinyMCE editor's settings so that it can handle full Unicode,
* verify if there are any other OCPL modules which need to be reconfigured to handle utf8mb4 (unlikely).

The alternative would be to migrate just the `cache_logs.text` and `pictures.title` fields for now, but then the code may need to distinguish between the two charsets.

Note that this may be an OKAPI-only issue. The OCPL editor is configured to simply not accept full Unicode, so there is nothing lost when submitting/processing the editor input. If there is any other OCPL data import (e.g. GPX, or Garmin field notes) which may contain Full-Unicode, then it would be an OCPL issue.

following5:
> The alternative would be to migrate just the cache_logs.text and pictures.title fields

alternative for point 2. All the other things would need to be done anyway.

wrygiel:
Partically merged in c717ee9.

Issue #408, opened 2016-04-03
Sending emails to users and reporting geocache problems


following5:
I will push a proposal for the interface of two new methods:
* services/users/send_email
* services/caches/report_problem (cmp. #393), which partiall builds on the email method.

Some things may need further discussion:

* I have proposed to document the **email logging** of the OC sites (table `email_user`). OCPL code additionally logs the email **subjects**, which I consider a data privacy no-go. I strongly suggest that OKAPI won't do that. If this will be implemented though, the docs need to be changed.

* Do we need to allow a **spam protection** for emails? The OC site's have none, and there is no spam problem. I think a simple solution would be to restrict the frequency of emails send by a user through OKAPI. This can be controlled via the email log. Otherwise, we probably would need to implement some complex protocol. (In OC.de wiki, asking for the first finder of geocaches has proven to work perfectly.)

* I have proposed to start with the three common **report_problem reasons** of the OC branches. (They are static and have internal IDs.) There are one / two additional report reasons at OCPL / OCDE, but allowing them may be expensive - I think it would similar definitions like the geocache attributes. Would like to postpone this.

We may also discuss this per Skype if it's too inconvenient here. ;)

following5:
(The report reasons are not important - they are just displayed to the OC admins like the report comment.)

following5:
See #409 for implementation.

following5:
Another email issue is localizing email templates, e.g. for the reports. They should be localized in the _recipient's_ language, but only OCDE directly allows that. For OCPL, we could copy the site's behaviour (use the _sender's_ language), or try to guess something from the (optional) country setting in the user's profile, from the most frequent country of logged geocaches and from Email TLDs. That's how Opencaching.DE `user`.`language` was initialized half a year ago.

following5:
I have proposed two additional boolean flags which can be queried for users:
* email_problems
* disclose_email_sender_address

Both are needed to fully resemble the OCDE "send email" function via OKAPI: It shows the "has email problems" info, and has a user configurable preset for disclosing the email address, which serves as default in the email form.

wrygiel:
I believe these are two separate issues: 1. emailing users, 2. reporting geocache problems; At first glance, I cannot understand why you choose to commit them together.

> We may also discuss this per Skype if it's too inconvenient here

You mean audio? If yes, then I agree! Talking is always faster that typing.

wrygiel:
In fact, I think we should talk *much* more :)

following5:
[...]

PR #409, opened 2016-04-03
New methods users/send_email and caches/report_problem, updates #408

following5:


following5:
will seprate it into two branches

PR #418, opened 2016-04-22
Fix #417: Get rid of ORIGIN_URLs

### LINKED ISSUE #417: Deprecate ORIGIN_URL
Originally, `ORIGIN_URL` served as an identifier of the source of the OC site. `OC_NODE_ID` was supposed to be equal to `4` for all development installations, so it couldn't serve this purpose.

Later, `ORIGIN_URL` also served to identify the set of attributes used by this installation. For example, OCNL used OCPL's `ORIGIN_URL` because it had exactly the same attributes.

However, all this causes much confusion. I propose some changes:

* OKAPI's `OC_NODE_ID` should never equal `4`. It should always refer to the OC NODE from which the database and code has been taken.

* `ORIGIN_URL` will be deprecated. We will use a normalized site name for matching attribute schemas.

wrygiel:


wrygiel:
@following5, @andrixnet

Issue #343, opened 2015-08-21
Alternative use of https


following5:
Opencaching.de is going to implement secure access. It is planned to enable parallel use of both https://www.opencaching.de and http://www.opencaching.de so that non-SSL-enabled or even non-redirect-enabled clients will not be broken.

OKAPI may be handled independently from this, but it would be nice to have it run with https, too. This would mean that SITE_URL would be set to either of both URLs. At a first glance, it looks like OKAPI can handle this, but I guess that it has not been fully tested in such an environment?

Are there any full URLs including the protocol stored in OKAPI tables or files?

wrygiel:
I don't think that the code requires any changes to support SSL. But the documentation might. Developers might wonder if they should pick HTTP or HTTPS scheme - pros and cons.

following5:
The usage of the SITE_URL setting needs verification. Generally, I think it should reflect the protocol used in the current request, so that any OKAPI links and redirects stay within the same protocol. At first glance I found one minor issue with that: [here](https://github.com/opencaching/okapi/blob/9019be4a147b49b4b061433365f702eb427e3a98/okapi/services/apisrv/installations.php#L69) the protocol in SITE_URL should be replaced by 'http', to be consistent with the installations.xml file.

wrygiel:
Perhaps we need to split it into two dynamically calculated variables - SITE_ID and SITE_HTTPX_URL? At the same time, the SITE_URL settings would be made private to avoid misuse.

BTW, I will be offline for the most of the next week.

following5:
> Perhaps we need to split it into two dynamically calculated variables - SITE_ID and SITE_HTTPX_URL? At the same time, the SITE_URL settings would be made private to avoid misuse.

Agree. I can do that.

I have scanned all direct and indirect uses of SITE_URL and found the following special cases; line numbers refer to [ver. 1113](https://github.com/opencaching/okapi/tree/72af34c59dfe1a44a3b0a0e85b688d41bcabea80):

**Notice about adding cron5 to the crontab**
* cronjobs.php, line 458
* update.php, 363

I think this should show always show _http:..._.

**Installations**
* installation.php, lines 26 and 27
* installations.php, lines 69 and 71
* menu.inc.php, line 73

I think this should always reflecte the _site ID_, which always starts with 'http:'.

* installations_box.tpl.php, line 5
* introduction.tpl.php, line 48

[...]

PR #419, opened 2016-04-22
Fix #416: Multiple base urls

### LINKED ISSUE #416: Support both www and no-www prefix for OAuth access?
I'm not yet sure how's that possible, but both www and no-www domains were (are?) accepted when making OAuth requests to OCPL's OKAPI. It wasn't possible when I first designed the system, but [seems to be possible now](https://github.com/opencaching/opencaching-pl/issues/513).

Since some apps (c:geo, for starters) are actually using the www prefix, it should also keep working this way. We need to analyze the code behind this.

wrygiel:


wrygiel:
@following5, @andrixnet, @deg-pl

okapi/core.php
```diff
@@ -1206,6 +1206,83 @@ public static function get_node_code()
...
+                    "http://www.opencaching.pl/okapi/",
+                );
+                break;
+            case 'OCDE':
+                $urls = array(
+                    "http://www.opencaching.de/okapi/",
+                    "https://www.opencaching.de/okapi/",
+                    "http://www.opencaching.it/okapi/",
+                    "http://www.opencaching.fr/okapi/",
+                    "http://www.opencachingspain.es/okapi/",
```
wrygiel: @following5 - does this seem correct?

okapi/core.php
```diff
@@ -1206,6 +1206,83 @@ public static function get_node_code()
...
+        /* Currently, there are no config settings which would let us allow
+         * to determine the proper values for this list. So, we need to have it
+         * hardcoded. (Perhaps we should move this to etc/installations.xml?
+         * But this wouldn't be efficient...) */
+
+        switch (self::get_node_code()) {
+            case 'OCPL':
+                $urls = array(
+                    "http://opencaching.pl/okapi/",
+                    "http://www.opencaching.pl/okapi/",
```
wrygiel: @deg-pl, @andrixnet  - OCPL will need keep allowing both of these domains when clients are accessing OKAPI - no Apache redirection should be allowed for `/okapi/` path.

deg-pl: It is possible. No problem ;)

okapi/core.php
following5: The only well-defined URL for accessing Opencaching.de OKAPI currently is http://www.opencaching.de/okapi/. OKAPI https is broken and will need #343 to be fixed.

following5: What about develsites?

wrygiel: > The only well-defined URL for accessing Opencaching.de OKAPI currently is http://www.opencaching.de/okapi/. 

Are you sure? They seem to be working correctly - I have checked. https://www.opencaching.de/okapi/services/apisrv/installation

> What about develsites?

Their URLs are added dynamically. See `get_allowed_base_urls` method.

following5: Most things work correctly with https, some do undesirable things (e.g. return http:// image urls in https context). search/by_urls does not work at all, but that should be adjusted anyway to the allowed-URLs list.

wrygiel: So, do you prefer your OKAPI to allow requests with all these prefixes, or would your rather risk breaking some apps and allow `http://www.opencaching.de/okapi/` only?

wrygiel: (OKAPI will reject all https requests if I remove the https URL from this list.)

following5: I have answered in #416.

following5: > search/by_urls does not work at all,

Sry, I confused things here. search/by_urls does work well for https and for with/without "www" prefix. It does _not_ work for opencaching.it, opencaching.fr and opencachingspain.es, but that's a separate issue.

okapi/core.php
```diff
@@ -1206,6 +1206,83 @@ public static function get_node_code()
...
+     *
+     * This is the URL which we want all *new* client applications to use.
+     * OKAPI will suggest URLs with this prefix in various context, e.g. in all
+     * the dynamically generated docs.
+     *
+     * Also see `get_allowed_base_urls` method.
+     */
+    public static function get_recommended_base_url()
+    {
+        return Settings::get('SITE_URL')."okapi/";
```
following5: An https url needs to be added here, so that https works on develsites.

wrygiel: I'm not sure why, but it seems that OCDE seems to be adding it dynamically. Perhaps you're feeding it directly to `SITE_URL` in your `okapi_settings.php`?

following5: Oops, you're right: https://www.opencaching.de/okapi/services/apisrv/installation

I thought that SITE_URL was not allowed to change from http://www.opencaching.de, but this was not enforced (because #343 was not done before starting https). Then, OKAPI will behave much better with https than I thought.

wrygiel: Why don't you return https *all* the time?

following5: See #343, postings 3, 4, 7, 9, 10. Unfortunately, #343 was not done before https was enabled at www.opencaching.de, and I forgot to disable https://www.opencaching.de/okapi. Therefore OKAPI-https now is in use, though it was not cleanly designed.

wrygiel: Aah, yes. I forgot about #343. We will probably have HTTPS enabled on OCPL soon, so I will need to look into it too.

andrixnet:
And these : 
https://github.com/opencaching/opencaching-pl/issues/105#issuecomment-214279481
https://github.com/opencaching/opencaching-pl/issues/105#issuecomment-214719373

PR #422, opened 2016-04-24
added 'log_uuids' field to logs/submit result; closes #403

### LINKED ISSUE #403: Return all uuids when submitting a log
logs/entry/submit may create two log entries, but returns just one of the uuids. It may be useful for developers to know both of them.

In the future, even three log entries may be created. E.g. if an OCPL team member submits a "temporarily not available" log for a cache that urgently needs maintenance: Log types 'OC team comment', 'temporarily not available', 'needs maintenance'. At OCDE we already have such logs.

What about adding a `log_uuids` return value, as an array which can contain any number of logs? Or is empty in case of an error.

following5:


wrygiel:
Seems okay!

PR #427, opened 2016-05-10
remove not necessary use statements

teiling88:
During my refactor work on opencaching.de i found some not necessary use statements. In this pr i only removed this ones.


following5:
I think this change is generally ok, but would prefer that Wrygiel decides how to deal with it. I may do the testing if necessary.

teiling88:
In PHPStorm - you get an notice for not used use statements - it quit easy to test ;-)

following5:
I guess that PHPstorm also evaluates use statements in included files ...? For transparency and robustness, I would not rely on that but keep the duplicate `use`s in the main file. Include files may be restructured, and then something at another place can break without noticing it.

following5:
>  keep the duplicate uses in the main file

... as far as the namespaces are referenced there

wrygiel:
My current IDE doesn't notify me when `use`-es are missing, so removing them increases the chance of me accidentally committing a bug. I would switch to a better IDE, but I don't know one (and PHPStorm is not free). I think we should postpone merging this commit until all the developers have better IDEs.

teiling88:
You can get free PHPStorm Licenses for your Project: https://www.jetbrains.com/buy/opensource/?product=phpstorm
opencaching.de already as get some licenses for free.
I removed only the not used use cases phpstorm mentioned.

wrygiel:
Thanks. I had a look at the form. It seems there are some requirements we would need to satisfy first for OKAPI to be able to acquire it (e.g. a licence URL). But perhaps I will try one day :)

wrygiel:
BTW, I think we could make use of Travis CI and some static code analyzer. I would very much like to experiment with GitHub's continuous integration options.

teiling88:
This sounds good can i help you?

wrygiel:
It was just a note to self :)

teiling88:
scrutinizer makes good work if you want to look, visit our repo :-) https://github.com/OpencachingDeutschland/oc-server3

wrygiel:
Looks promising. Thanks! I hope I will find the time to explore this option.

PR #461, opened 2017-04-19
house keeping

teiling88:
- optimize require_once statements
- add missing phpdoc params
- adding missing method access modifier (important for php 7.0)

### DIFF
--- a/okapi/controller.php
+++ b/okapi/controller.php
@@ -14,12 +14,11 @@ use Exception;
 
 $GLOBALS['rootpath'] = '../'; # this is for OC-code compatibility
 
-require_once($GLOBALS['rootpath'].'okapi/core.php');
+require_once __DIR__ . '/core.php';
 OkapiErrorHandler::$treat_notices_as_errors = true;
-require_once($GLOBALS['rootpath'].'okapi/urls.php');
+require_once __DIR__ . '/urls.php';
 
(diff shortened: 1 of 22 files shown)


wrygiel:
> optimize require_once statements

Merged! Thanks!

> add missing phpdoc params

I think we should skip this one. I find such automatically generated docs more vague than no docs at all. Most are empty, formatting was broken in many places, and the "automatically guessed" datatypes often don't match the actual datatypes expected/returned.

> adding missing method access modifier (important for php 7.0)

I tried to find some information on this. Can you tell me why is it important in PHP 7?

wrygiel:
> > adding missing method access modifier (important for php 7.0)
> 
> I tried to find some information on this. Can you tell me why is it important in PHP 7?

Merged it too (but I still wonder why this is important).

wrygiel:
BTW, you missed one slash. I got this in logs:

```
PHP Fatal error:  okapi\\Settings::default_gettext_init(): Failed opening required '/srv/ocpl/okapilocale/locales.php'
```

Fixed in 7f9127f96c704fa30ac547d33339ccb5b68297f3.

teiling88:
thank you for your fast response! 

> Can you tell me why is it important in PHP 7?
my fault - it is a should not a must definition :-(

PR #466, opened 2017-05-31
unify require_once statements

teiling88:


wrygiel:
Can you explain why?

wrygiel:
Personally, I find this:

```php
require_once($GLOBALS['rootpath'].'okapi/service_runner.php');
```

much more understandable than this:

```php
require_once __DIR__ . '/../../service_runner.php';
```

teiling88:
With require_once __DIR__ modern IDE`s can check the path if it is correct and supports auto completion. Furthermore you can click on oc_session.php to jump directly in this file. 

Example from PHP-Storm
image

wrygiel:
Perhaps it makes it better for your IDE, but it makes it worse for mine :)

I have an alternative idea: How about we agree on a common [include path](http://php.net/manual/en/function.set-include-path.php)? If we do, then `require_once "okapi/lib/oc_session.php"` will work for both of us.

wrygiel:
Check out version 3a1eae125d2f5bca2fa3b0d394ee3c64037b97ce. First add your `rootpath` to [your IDE's include paths](https://www.jetbrains.com/help/phpstorm/2017.1/configuring-include-paths.html).

teiling88:
> If we do, then require_once "okapi/lib/oc_session.php" will work for both of us.

This solution works but creates possible side effects. It is not guaranteed that the right file is loaded if this path appears in more then one include_path. 

In PhpStorm with the EA Inspection Module the following advise appears: "This relies on include_path and not guaranteed to load the right file. Concatenate with `__DIR__` or use namespaces + class loading instead."

If it is ok for you i would create an pull request für the case "namespace + class loading". In this pr  I would gladly change the namespace and class structure to the [PSR-4](http://www.php-fig.org/psr/psr-4/)-Standard.

wrygiel:
> If it is ok for you i would create an pull request for the case "namespace + class loading". In this pr I would gladly change the namespace and class structure to the PSR-4-Standard.

Okay, let's do that :)

I will also try to find some time to read about PSR-4, so I can review your commits better. BTW, please try to commit your changes in separate smaller commits, so it's easier to review and test them.

teiling88:
> Okay, let's do that :)

great :+1:

I started yesterday with a short implementation (not tested and not complet) but I tried to create small commits. You can see in this branch:  
https://github.com/teiling88/okapi/tree/feature/adding-psr-4

Tomorrow I will be on vacation for 14 days so I will finish this branch after that. :-)

wrygiel:
I think we need a new issue for this topic.

PR #467, opened 2017-06-06
Feature/adding psr 4

teiling88:
This is an example PSR-4 Pull Request. I "only" changed the service directory to proper PSR-4 standard. I tested overview function and so on. I **didn't** test everything caused of my vacation. 

I hope with this Pull Request the PSR-4 Structure is a little bit clearer. To avoid braking changes i changed no Namepsace - i only renamed the files and moved them to the correct PSR-4 Space ;-) 

In this stage of development I only used composer for autoloading and I commited the generated autoloading files. In my opinion we can remove this in a further development version but for the beginning it  should be a little bit easier :)


wrygiel:
Haha, I see you took the message about "small commits" to heart. :) Not exactly what I meant, but it's okay ;)

I have reviewed your changes and now I can understand what you want to achieve. I like the idea!

I have also tried to run this new version, and met with errors. I fixed those which I found, but I expect there might be much more (because I have tested only two methods). I will publish a branch with my changes soon - **please merge them with yours once you're back** (so that I will have a new base for review).

I have reverted some of your codestyle changes. I have nothing against changing OKAPI's codestyle, but:

* Such changes should be done globally and simultaneously (as opposed to making them just locally, when the opportunity arises).
* If we do, we should first discuss it (OKAPI's current code style is described [here](https://github.com/opencaching/okapi/blob/affdbbb476ba3ebd58a6b05cea1eaeb2e3640db9/etc/CODESTYLE.md)).
* We should not mix codestyle changes with this PR. We need a separate issue for that (and a separate branch too).

wrygiel:
> I will publish a branch with my changes soon - **please merge them with yours once you're back** (so that I will have a new base for review).

Just noticed that I have write-permission to your fork. So I pushed my changes directly there.

teiling88:
Hey @wrygiel,
sorry for my long delay. I will continue this pr in the next week.

Greetings

wrygiel:
Okay!

teiling88:
Today, i wrote the first unit tests for the okapi integration in our de repo: 
https://github.com/OpencachingDeutschland/oc-server3/pull/503/commits/abfa15e1fd53bb7e1d9ce6d7471535592880e46b 
https://github.com/OpencachingDeutschland/oc-server3/pull/503/commits/8ec363ff013f5fecc2eaf0f53e890319c9beb879

If you can see, the testing thing ist quite simple and saves in the future much testing time :-) 
Tomorrow the test should pass on travis-ci too :-D

wrygiel:
I'm a big fan of unit-testing myself. (OKAPI doesn't have any though, because I wasn't such a big testing enthusiast, when I first created it. And didn't have time to add them later on.)

teiling88:
Hey @wrygiel,

i checked all functions til authentication level 1 and all of them worked. 

I think - this pr is so far for a review from your side :-)

Greetings

PS: On a later step i will move the unit tests from our oc repo to here!

wrygiel:
You [wrote](https://github.com/opencaching/okapi/pull/467#issuecomment-315080702) that you wanted to continue your work on this, but I haven't seen any more commits in this PR, so I assumed you are still working on this - did I miss something?

wrygiel:
I will try to review next week!

teiling88:
> You wrote that you wanted to continue your work on this, but I haven't seen any more commits in this PR, so I assumed you are still working on this - did I miss something?

I created the unit tests in the OpencachingDe Repo. Have a look in this pr: https://github.com/OpencachingDeutschland/oc-server3/pull/503 after PSR-4 integration i will move the unit tests to the Okapi repo if you want?


> I will try to review next week!
Thank you!

wrygiel:
Yes, having some architecture for testing would be most welcome. Not yet sure how you intend for these tests to work though. Since OKAPI is supposed to work on two OC branches, and tests would need to have access to OC databases, then I guess that Travis will also need to create a test-database (or two of them - for both OCPL and OCDE structures) upon each run?

teiling88:
This sql dumps for travis exists allready for the de branch. So I would start with the de branch and after that we could finish pl testing together?

wrygiel:
BTW, I can see that you have rebased your changes. Why didn't you simply merge master to your PR?

This approach is safer, because one can clearly see what happened. When you rebase, I cannot be sure that you haven't changed something in some of the previous commits which I have already reviewed. :)

wrygiel:
Also, can you please respond to this comment?

> I have also tried to run this new version, and met with errors. I fixed those which I found, but I expect there might be much more (because I have tested only two methods).

Since I got quite a lot of errors on the last review (and I have tested only two methods), and I can see that you have changed only 3 files since then, therefore I am still kind of afraid that there are lots of errors in your PR. Can you tell me how exactly did you test it? Somehow I fell there's a big chance we will receive many errors just after we put it on production, and we will need to revert it.

wrygiel:
I have tested it again, and it still doesn't work. For example, when I try to access:

```
http://local.opencaching.pl/okapi/services/apiref/method_index
```

It keeps redirecting me to:

```
http://local.opencaching.pl/okapi/services/apiref/method_index/
```

And this results in an error:

image

I cannot merge it.

wrygiel:
I think you need to perform some more testing before this is production-ready. :)

teiling88:
Nice issue :D During our new directory structure, the apache service thinks he should add an trailing slash because the api path exists as a directory. So the option DirectorySlash Off in the .htaccess file fixes this issue. 

This was already deactivated on my development vm :-\

I removed the set_include_path to in the last commit.

teiling88:
> I think you need to perform some more testing before this is production-ready. :)

So this okapi version is at the moment online on opencaching.de

teiling88:
> BTW, I can see that you have rebased your changes. Why didn't you simply merge master to your PR?

> This approach is safer, because one can clearly see what happened. When you rebase, I cannot be sure that you haven't changed something in some of the previous commits which I have already reviewed. :)

ohh - this is our standard development process for updating a feature branch :-\

okapi/OkapiScriptEntryPointController.php
```diff
@@ -20,17 +20,11 @@
 # variable is being set up by the OC site. If it is called via the controller
 # endpoint (this one!), then we need to set it up ourselves.
 #
+error_reporting(E_ALL);
+ini_set('display_errors', 'on');
```
wrygiel: And you are running this in production?

okapi/core.php
```diff
@@ -28,7 +28,7 @@
 function get_admin_emails()
 {
     $emails = array();
-    if (class_exists("okapi\\Settings"))
+    if (class_exists(Settings::class))
```
wrygiel: This change doesn't make much sense, I think. The purpose of this line seems to be to check if the class exists. Previously, if this class didn't exist, `class_exists` would return `false`. After your change, I think that an exception will be thrown, even before `class_exists` is called.

okapi/OkapiScriptEntryPointController.php
teiling88: Yes, I looked for the error_log - now entries :-)

okapi/core.php
wrygiel: In other words, in your version, the admin will never receive an email if parsing of `Settings` fails. In the original implementation - he would.

teiling88: i checked it with the following snippet:

```php
<?php
namespace okapi;

var_dump(class_exists(Settings::class));
```
via https://3v4l.org
It works on all php versions

wrygiel: It seems that `Settings:class` evaluates to a string... PHP is awsome ;)

```php
<?php
namespace something;
print Settings::class;
```

teiling88: Yeah php allows so much correct version of the same result :-D

wrygiel:
Ok, I ran some manual tests, and couldn't break it. I'll give it a try on production.

@opencaching/opencaching-pl-lead-programmers. This is a major change, so if something breaks, and I seem to be offline (I plan to be, starting tomorrow, for a couple of days), then please revert the commit.

teiling88:
> Ok, I ran some manual tests, and couldn't break it. I'll give it a try on production.

Thank you!

wrygiel:
Merged in 5bb93099fdf9282c8e240b0f5ad854d4b229a64f.

PR #469, opened 2017-06-13
Update installations site_urls to use HTTPS

prmtl:
HTTPS should be used whenever possible through the API.
API endpoint `okapi/services/apisrv/installation` is already
returning `site_url` with HTTPS.

### DIFF
--- a/etc/installations.xml
+++ b/etc/installations.xml
@@ -4,10 +4,10 @@
     read by every OKAPI installation, *directly from repository*. See the code for
     services/apisrv/installations method for explanation.
     -->
-    <installation site_url="http://opencaching.pl/"/>
-    <installation site_url="http://www.opencaching.de/"/>
-    <installation site_url="http://www.opencaching.us/"/>
+    <installation site_url="https://opencaching.pl/"/>
+    <installation site_url="https://www.opencaching.de/"/>
+    <installation site_url="https://www.opencaching.us/"/>
     <installation site_url="http://www.opencaching.nl/"/>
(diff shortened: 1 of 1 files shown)


prmtl:
A different way could be to return both urls: HTTP as site_url and HTTPS as site_url_secure.

wrygiel:
Did you test this change? Will this actually affect the URLs returned by OKAPI? Won't this break older OKAPI installations?

prmtl:
Yes, I tested it.

About breaking older installation: this will not break them. I'm only worried about older API clients that might have not support HTTPS or certificates used by websites - this is why I'm also thinking about adding "secure" version of those urls as another attribute returned by API (which is probably the safest method). What do you think?
But on the other hand `okapi/serces/apisrv/installation` already returns urls with HTTPS for some installations (like openchaching.pl)

wrygiel:
If I can follow all the outcomes properly, then your PR will only alter two fields (`site_url` and `okapi_base_url`) in a single method (`services/apisrv/installations`). Changing `site_url` cannot break anything, but changing `okapi_base_url`, in theory, might - if both of the following are true:

1. Someone chose to dynamically fetch the list of OKAPI installations from this method and connect to these installations dynamically (without a Consumer Key). Such connections won't yield much data, because Level 0 Authentication would need to be used.

2. The connection procedures he had used expect - for some reason - the given URL to be `http://`.

However, I think that the chance of both of these events occurring is pretty slim. We might risk it.

Issue #482, opened 2017-08-18
Documentation policy regarding branch-specific features


following5:
In the documentation of services/cache/geocache, two "OCPL" icons have popped up, marking fields that are only filles with content by OCPL code. This diverts from the OKAPI documentation policy not to be specific about which code fork supports which features.

There are also features only available at OCDE. It would be consistent to mark them as "OCDE", too. How shall we go on here? Has the policy changed and docs may now be specific about branch-dependend features?

wrygiel:
Yes, there has been some pressure from OCPL developers who wanted to add support for OCPL-specific features (which would always return empty placeholders in OCDE). These features have not yet been added, but I have added the icons to make them possible.

BTW, there also are some features marked with "OCDE" icon.

wrygiel:
Of course, the policy doesn't change - all methods still must work on all forks. However some methods work "much better" on some forks. This is what the icons indicate.

following5:
> However some methods work "much better" on some forks.

Yes, that's what I meant. In the past, the docs were unspecific in such cases and only stated that "some installations" may behave in this or that way. The change now is to explicitly say which code branches work how.

So it should be fine to add an OCDE symbol for similer cases, where methods work "better" on OCDE ...?

wrygiel:
> So it should be fine to add an OCDE symbol for similer cases, where methods work "better" on OCDE ...?

Yes.

PR #483, opened 2017-08-18
added some docs on branch-specific behaviour; updates #482

following5:


following5:
directly committed to master

PR #484, opened 2017-08-19
updated changelog and fixed github commit links

following5:


following5:
directly commited to master

PR #486, opened 2017-08-21
misc - fix some auto loading issues

teiling88:


okapi/views/devel/cronreport.php
```diff
@@ -18,7 +18,7 @@ public static function call()
         $response->content_type = "text/plain; charset=utf-8";
         ob_start();
 
-        require_once "okapi/cronjobs.php";
+        require_once __DIR__ . '/../../cronjobs.php';
```
wrygiel: We wanted to avoid such relative imports. Is it necessary?

okapi/views/devel/tilereport.php
```diff
@@ -13,7 +13,7 @@ public static function call()
     {
         Okapi::require_developer_cookie();
 
-        require_once "okapi/cronjobs.php";
+        require_once __DIR__ . '/../../cronjobs.php';
```
wrygiel: Same here

okapi/views/update.php
```diff
@@ -11,7 +11,7 @@
 use okapi\services\replicate\ReplicateCommon;
 use okapi\Settings;
 
-require_once "okapi/cronjobs.php";
+require_once __DIR__ . '/../cronjobs.php';
```
wrygiel: and here

okapi/views/devel/cronreport.php
teiling88: Yes because we removed the set include path code. We can remove such includes if we finish for all files the PSR-4 Structure.

wrygiel: > We can remove such includes if we finish for all files the PSR-4 Structure.

Oh, I thought you did :)

wrygiel: So now OKAPI is using some kind of a "hybrid structure"?

teiling88: services a done - views are under construction and in this step i will fix the rest classes.

wrygiel: Great :)

Issue #411, opened 2016-04-05
Reporting geocache problems


following5:
Spin-off if issue  #408.

What may need further discussion are the "report reaons". This is an internal-ID field in the geocaches report table. Theres are the available reasons for OCPL

* Cache on private property
* **Needs archiving**
* Copyright violation
* Other

and for OCDE:

* **Cache is gone**
* Cache on private property
* Copyright violation
* **Description is unusable**
* Other

I suggest to start with the common reasons, because providing the different reason sets will be expensive, similar to the attributes definitions. "Other" will do well.

following5:
This is the proposed definition of caches/report_problem:

https://github.com/following5/okapi/tree/feature/report

following5:
There is a request from the Opencaching.DE support team, which could be resolved by implementing report functionality in OKAPI. I will submit an interface proposal for these new methods as a Pull Request:

- service/caches/report/submit:  submit a new geocache report
- service/caches/report/status:  query the status of a one or more previously submitted reports
- service/caches/report/reports: query list of pending reports for a geocache

My proposal is to implement the common OCDE/OCPL "report to Opencaching team" functionality. The OCPL-specific functionality "report to the owner" - which would need email sending through OKAPI - could be added later.

following5:
Added another proposed method service/caches/reports/reasons to cope for the different available report reasons in OCPL and OCDE code.

wrygiel:
I read this thread *after* I reviewed to pull-request :) Wrong order of emails :)

wrygiel:
Anyway, I didn't even know that such functionality is available in OCPL. Good to know!

deg-pl:
Maybe we can sync report reasons on OC DE and OC PL sites.
"Needs archiving" and "Cache is gone" - I think they are very similar reasons. Maybe we can chose something common for all sites?
"Description is unusable" - I'm nearly sure, that OC PL sites can add this reason.
We have small problem with "Cache on private property" because polish meaning is wider - something like "incorrect location".

following5:
I think it will be difficult to sync it.

E.g. "needs archiving" doesn't make sense at OC.de: Reporters here describe what is wrong with the cache, and then the support team desides to
* just inform the user, or
* log "cache needs maintenance" or "listing is outdated", or
* disable, archive, lock or lock/hide the cache.

Also, "needs archiving" is somewaht redundant to the other reasons, which may also require archiving.

So what could be a replacement for "needs archiving" at OCPL?

[...]

PR #487, opened 2017-08-23
proposal of new cache reporting methods; updates #411

following5:


following5:
needs further discussion with Opencaching.de support team

okapi/services/caches/reports/reasons/docs.xml
```diff
@@ -0,0 +1,32 @@
+<xml>
+    <brief>Get list of allowed geocache report reasons</brief>
```
wrygiel: Missing "the" before "list", I think? (similar in other methods)

okapi/services/caches/reports/status/docs.xml
```diff
@@ -0,0 +1,40 @@
+<xml>
+    <brief>Query the status of geocache reports</brief>
+    <issue-id>TODO</issue-id>
+    <desc>
+        <p>This method queries the status of one or more geocache reports that were
+        submitted trough <a href="submit.html">services/caches/reports/submit</a>.</p>
```
wrygiel: I think there was a plugin for generating values for hrefs (which takes the full path of the method and returns a valid URL). It's better for future-compatibility.

okapi/services/caches/reports/status/docs.xml
```diff
@@ -0,0 +1,40 @@
...
+            <li>
+                <b>null</b> - there is no report with this ID, or
+            </li>
+            <li>
+                <p>a dictionary, having this structure:</p>
+                <ul>
+                    <li>
+                        <p><b>status</b> - the status of the report, which is
+                        either <b>Unassigned</b>, <b>In progress</b> or <b>Closed</b>,</p>
+                    </li>
```
wrygiel: Are you sure there won't be any more statuses in the future? If there will be, a new field will be introduced, yes?

okapi/services/caches/reports/status/docs.xml
```diff
@@ -0,0 +1,40 @@
...
+                    <li>
+                        <p><b>admin_id</b> - if the report is in progress, the user ID
```
wrygiel: `assignee_uuid`? (we use UUIDs as primary keys for users in all other methods)

okapi/services/caches/reports/status/docs.xml
```diff
@@ -0,0 +1,40 @@
...
+                        of the Opencaching team member to whom it is assigned;
+                        <b>null</b> if the report is not in progress,</p>
+                    </li>
+                    <li>
+                        <p><b>admin_name</b> - the team member's username, or <b>null</b>
+                        if the report is not in progress.</p>
```
wrygiel: I think it would be better to remove both `admin_id` and `admin_name`, and replace them with `assignee` object (which in turn would have the same structure as the objects returned in `services/users/user`).

If it's a problem for you to create an object here, then at least we should name these two fields the same way as they are named in the `user` object, that is (`assignee_uuid` and `assignee_username` - to be deprecated in the future, and replaced by `assignee`).

okapi/services/caches/reports/submit/docs.xml
```diff
@@ -0,0 +1,72 @@
...
+        non-urgent cases, before reporting to the OC team. The owner currently
+        cannot be contacted via OKAPI but only through the OC website.
+        You may link to the owner's profile page as given
+        by <a href="../../users/user.html">services/users/user</a>;
+        there is an email button.</p>
+    </req>
+    <req name='reason'>
+        <p>String - the reason for the report; any of the reason values as returned
+        by <a href="reasons.html">services/caches/reports/reasons</a>.</p>
+        <p>It is recommended to let the user chose out of a list of all allowed
```
wrygiel: "choose"?

okapi/services/caches/reports/reports/docs.xml
```diff
@@ -0,0 +1,15 @@
+<xml>
```
wrygiel: I believe none of these will work on OCPL, right? So we need some of these newly introduced "tags" here.

okapi/services/caches/reports/status/docs.xml
following5: ok, it's %OKAPI:methodargref

okapi/services/caches/reports/status/docs.xml
following5: have removed this field

okapi/services/caches/reports/status/docs.xml
following5: have removed this field

okapi/services/caches/reports/reports/docs.xml
following5: It all works at both OCPL and OCPL

okapi/services/caches/reports/status/docs.xml
following5: Actually, there is also the status "Look here!" at OCPL, which cannot unambiguously be mapped to "Unassigned" or "In progress".  To keep things simple, I have now merged all three to "Open".

I think it's very unlikely that a new status will be added which cannot be mapped to either "Open" or "Closed". If so, we would indeed need a status2 field.

following5:
(rebased, the first commit - c7eac46 - is identical to the old commit reviewed by wrygiel)

following5:
acidentially merged into master, have revertet that and will issue a new PR, sorry

following5:
I accidentially merged this feature-under-development into into master. This has created a new field `reports.uuid` in the database of all OCPL sites (and sent some error messages to admins).

Have reverted the merge, but suggest to keep the `reports.uuid` field for now, instead of removing it and then readding it again when this feature is deployed. You can safely ignore it; it will be initialized later.

deg-pl:
@following5 reports.uuid has no default value and not permit null value. 
We don't generate uuids while user creates report - so this breaks strict mode in MySQL.
In strict mode you MUST insert all fields into row, which has no default values. Please correct reports.uuid structure - add default value or/and allow null value.

following5:
> Please correct reports.uuid structure - add default value or/and allow null value.

done

If this change eventually goes online, I would add code to `reportcache.php` which inserts the uuids. If it won't go online, I would take care to renove the uuid field. For now it's just a dummy.

deg-pl:
Now is OK. Thanks!

PR #489, opened 2017-08-24
misc/adding psr4 part 2

teiling88:
This is PSR-4 Refactoring Part 2 and the last one of it. This PR is at the moment for your information.

* View/Documentation is ready

I have the following open todos:
* check services
* check cronjobs

If I'm ready with testing I will inform you.


teiling88:
So i tested the okapi on oc.de. Write logs etc worked fine. 

I will have a look on the cronjobs but so far as I can see everything works fine :-)

teiling88:
The PR is since 6 days live on opencaching.de. Everything works fine. 

No error message appeared in the log file :-)

following5:
The new top directory structure looks a bit cluttered:
* Consumer
* CronJob
* Exception
* lib
* locale
* OAuth
* Request
* Response
* services
* static
* Token
* vendor
* views

Maybe all the "core" OKAPI classes can be moved to a "core" subtree, including thos that currently reside in the top okapi dir?

following5:
This change creates the files "okapi/okapi.php" and "okapi/Okapi.php", which differ only in case. Could you resolve that? Tried to test it on my develsite, but it won't run because it resides on an NTFS file system (case insensitive).

teiling88:
> Maybe all the "core" OKAPI classes can be moved to a "core" subtree, including those that currently reside in the top okapi dir?

This would create a breaking change - cause you have to add the new subnamespace core.

In my opinion this is much friendlier to read than several files with thousands of lines.

> Could you resolve that?

done!

following5:
Thanks. Now it runs on my develsites; will do some OCPL testing.

> This would create a breaking change - cause you have to add the new subnamespace core.

Well, this change already adds the subnamespaces "Consumer", "CronJob", "Exception" etc., right? So inserting a "core" inbetween would be a similar change and should not break anything ...?

But this is not critical anway, may be done in a second refactoring step.

teiling88:
> Well, this change already adds the subnamespaces "Consumer", "CronJob", "Exception" etc., right?

No. The namespaces like cronjob already exists.

> But this is not critical anway, might be done in a second refactoring step.

I would prefer this :+1:

following5:
> No. The namespaces like cronjob already exists.

Ok, 'cronjob' and 'oauth' already exist (and are renamed). Consumer, Exception, Request, Response and Token are added.

So far I found an issue with update: It throws an exception in both OCDE and OCPL environment, both when called via  http://&lt;develsite>/okapi/update or Facade::database_update():

<pre>
Current OKAPI database version: 118
Oops... Something went wrong on *our* part.

Message was passed on to the site administrators. We'll try to fix it.
Contact the developers if you think you can help!

BUT! Since the DEBUG flag is on, then you probably ARE a developer yourself.
Let's cut to the chase then:

===== ERROR MESSAGE =====
Exception:

=========================

--- Stack trace ---
#0 /mnt/ocpl/okapi/views/update/View.php(61): okapi\views\update\View::_call()
#1 [internal function]: okapi\views\update\View::call()
#2 /mnt/ocpl/okapi/lib/OkapiScriptEntryPointController.php(52): call_user_func_array(Array, Array)
#3 /mnt/ocpl/okapi/index.php(46): okapi\lib\OkapiScriptEntryPointController::dispatch_request('/okapi/update')
#4 {main}

following5:
Oops, that's a local problem not related to this change. (local DB version is one step ahead of OKAPI master)

following5:
Okay, this code looks stable. As this PR blocks OCDE development and after being back on the [OKAPI spamlist](https://github.com/opencaching/opencaching-pl/commit/be5e222af4fa2101eec3384a1f12101afcdf4a50), let's give it a try.

following5:
Looks like this merge has broken OKAPI deployment. Neither http://rygielski.net/r/okapi-latest nor OCPL repo has been updated. Three hours before everything was working fine.

It happened that a single commit was eaten, but [four in a row](https://github.com/opencaching/okapi/commits/master) is unusual. Now we must wait for @wrygiel to fix it. :(

following5:
There is another OKAPI issue to be fixed. I will try to reanimate OKAPI deployment by reverting this merge.

Github does not allow to directly reopen this PR. If the revert does not reopen it, please issue a new PR!

teiling88:
does the deployment worked after the revert @following5 ?

following5:
Unfortunately - no.

I think that wrygiel will read this as soon as he is back and then care about re-adding your change. But to be sure that it's not overlooked, you may decide to open a new PR (just reverting my commit a8b062f1).

PR #491, opened 2017-08-29
added services/caches/edit for log passwords; resolves #488

### LINKED ISSUE #488: [Feature request] Change password in cache via OKAPI
Hello, after long break.

I want to prepare cache with password, but my idea is related with often changing of this password. 
Im looking for idea how to automate this process.

Maybe OKAPI should do things like that?

following5:


following5:
method href's in the docs here also need to be changed to %OKAPI:methodargref:...%

PR #499, opened 2017-09-05
PSR-4 Part 2 - second try

teiling88:


following5:
See #489 for previous discussion.

following5:
OKAPI deployment is back to life. But I don't dare to merge this change again, will leave it to wrygiel.

wrygiel:
Hi! I'm back from holidays, but I'm completely buried in work. I didn't even have time to read all emails I got since then, and I probably won't in the upcoming weeks, sorry about that. :( Please don't wait for my review and continue with your changes. I trust you to keep OKAPI safe and backward-compatible :)

> But I don't dare to merge this change again

From the sound of it, I suspect that something broke on production when I was offline. It happens sometimes. As long as you coordinate with @opencaching/opencaching-pl-admins (e.g. they will forward you OKAPI error emails), you should be able to pick up all mistakes soon enough. Perhaps @deg-pl could even add you to [the list of emails who receive OCPL OKAPI error messages](https://github.com/opencaching/opencaching-pl/blob/39b9356f02a991c2f524c097256f527579f6c9cc/okapi_settings.php#L39) (I just now see that you are already on this list).

following5:
> Please don't wait for my review and continue with your changes. I trust you to keep OKAPI safe and backward-compatible :)

Ok, thanks.

> From the sound of it, I suspect that something broke on production when I was offline.

http://rygielski.net/r/okapi-latest was not updated. Is this connected to the PL production site?

following5:
Okay then ... let's try again. If it breaks again, I may switch to manual OKAPI deployment. ;-)

following5:
Second try, second failure. OKAPI deployment is dead again.

Ok then, I will manually deploy to OCPL.

following5:
> Second try, second failure. OKAPI deployment is dead again.

Of course it is dead. It tries to insert the version numbers into core.php, but there no longer is a core.php. The numbers now are in Okapi.php.

And "facade.php" - which is included in OCPL code - also does no longer exist under this name. It's now Facade.php. Didn't notice on my test site because the file system is case-insensitive.

following5:
Will fix this later today in postcommit.py.

following5:
> Will fix this later today in postcommit.py.

... but it MUST NOT be fixed before merge of https://github.com/opencaching/opencaching-pl/pull/1186, because OCPL code must be changed synchronously! See https://github.com/opencaching/opencaching-pl/pull/1186/commits/c369c0a128c3d3c20f56884dc8dd02819819aba2.

Issue #344, opened 2015-09-05
Introduce log_uuids somewhere in generated GPX files


MavEtJu:
When I submit a log through http://www.opencaching.nl/okapi/services/logs/submit for example, it returns a log_uuid. However, that log_uuid is not reflected in the groundspeak:log id returned in the GPX data.

Returned data:
data: {"success":true,"message":"Your cache log entry was posted successfully.","log_uuid":"3369a85c-cd83-4e24-b3f7-ea733c4a8bb1"}

Later on retrieved GPX file:
<groundspeak:log id="4139">

It would be nice if that number 4139 could be returned by the services/logs/submit function.

Edwin

following5:
You may use the services/logs/entry method to retrieve the ID number which will be returned in the gpx data:

http://www.opencaching.nl/okapi/services/logs/entry?log_uuid=3369a85c-cd83-4e24-b3f7-ea733c4a8bb1&fields=internal_id&consumer_key=...

wrygiel:
This inconsistency is because log IDs in Groundspeaks GPX file need to be integers. We would need to introduce a custom OKAPI XML namespace in the GPX file in order to be able to include UUIDs there. But I'm not sure if anyone would use such extra namespace.

following5:
I like the idea of a custom _Opencaching_ namespace in GPX files, to transport all the OC information which is missing there. This should be a joined effort of OCPL, OCDE and OKAPI developers.

wrygiel:
I was thinking about it when I was making the GPX formatter, but would any client libraries use it? Might be, that would just make our GPX files bigger, for no real reason.

following5:
I think that GSAK may use it. It's a very common tool and already supports the gc_ocde:attrs OConly attribute.

following5:
Meanwhile there is a request at OCDE to include a flag for "log password needed" into GPX files, which then could be evlauated by c:geo.

I will do a sample implementation for an 'opencaching' GPX extension.

following5:
@MavEtJu
The log UUIDs now can be included in GPX files. Set `latest_logs=true` and `ns_oc=true`. You will find the UUIDs in 

```
<wpt>
    <oc:cache>
        <oc:logs>
            <oc:log id="<id>" uuid="<uuid>"
            </oc:log>
        </oc:logs>
    </oc:cache>
</wpt>
```

PR #511, opened 2017-09-14
Proposal of a Common Opencaching GPX extension, see #344

following5:


following5:
@teiling88 
This is a proposal for an own Opencaching GPX extension, which would allow to put Opencaching-specific information into GPX files. It starts with an OKAPI implementation of log uuids (OKAPI issue #344) and a "requires password" field for caches (see [OC.de forum topic](https://forum.opencaching.de/index.php?topic=4787)). The latter one would allow to show password requirement e.g. in c:geo, when the cache was imported via GPX.

I suggest do deposit the XML schema definition at https://opencaching.pl/okapi/static/ocgpx. So it would be maintained in the OKAPI project, which ensures stability and compatibility across all OC sites and versions, while the reference file is at the PL site which is always most current due to automatic OKAPI deployment.

Would you support this OC GPX extension on your websites, i.e. in native GPX generation in OCPL / OCDE code? I may also do the implementations for that.

Do you have any comments or suggestions? I am not experienced in creating XML Schemas, just collected the information needed from tutorials and from existing GPX extensions.

following5:
Example output:

<pre>
&lt;gpx
...
    xmlns:oc="https://opencaching.pl/okapi/static/ocgpx/1"
    xsi:schemaLocation="
...
        https://opencaching.pl/okapi/static/ocgpx/1
        https://opencaching.pl/okapi/static/ocgpx/1/ocgpx.xsd
>
&lt;wpt lat="45.020567" lon="7.037033">
    &lt;groundspeak:cache archived="False" available="True" id="1">
...
        &lt;groundspeak:log id="33">
...
            &lt;oc:log>
                &lt;oc:uuid>426A3DF7-CF33-D83F-EB68-63711932E09F&lt;/oc:uuid>
            &lt;/oc:log>
        &lt;/groundspeak:log>
    &lt;/groundspeak:cache>
    &lt;oc:cache>
        &lt;oc:req_passwd>true&lt;/req_passwd>
    &lt;/oc:cache>
</wpt>
</pre>

following5:
@rsudev, @SammysHP:
Would you support this in c:geo? E.g. evaluate the OC `req_passwd` flag from GPX files?

okapi/static/ocgpx/1/ocgpx.xsd
```diff
@@ -0,0 +1,30 @@
...
+        <xsd:documentation>
+            Common Opencaching extensions to the Topografix GPX schema.
+        </xsd:documentation>
+    </xsd:annotation>
+
+    <xsd:element name="cache" type="cache_type" />
+
+    <xsd:complexType name="cache_type">
+        <xsd:sequence>
+            <xsd:element name="logpw" type="xsd:boolean" minOccurs="1" maxOccurs="1" />
```
following5: This must be renamed to "req_passwd".

okapi/static/ocgpx/1/ocgpx.xsd
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xsd:schema
+    xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+    xmlns="https://opencaching.pl/okapi/static/ocgpx/1"
```
wrygiel: Perhaps it's safer to use xmlns in OKAPI's github namespace? See [here](https://github.com/erasmus-without-paper/ewp-specs-api-discovery/blob/v4.1.0/manifest-example.xml#L2) for example. GitHub URLs are highly unlikely to change, and we don't convey a bad impression that "this is for OCPL only".

I would even suggest to put this XSD outside of the OKAPI repository - we could make a separate repository solely for that single XSD (e.g. `https://github.com/opencaching/gpx-extension`), and introduce semantic versioning similar to the one I used in the example linked above (`stable-v1` branch). I have designed [tons of XSDs in this fashion](https://developers.erasmuswithoutpaper.eu/) in recent years and it seems to be working very well, so far.

okapi/static/ocgpx/1/ocgpx.xsd
```diff
@@ -0,0 +1,30 @@
...
+    targetNamespace="https://opencaching.pl/okapi/static/ocgpx/1"
+    elementFormDefault="qualified">
+
+    <xsd:annotation>
+        <xsd:documentation>
+            Common Opencaching extensions to the Topografix GPX schema.
+        </xsd:documentation>
+    </xsd:annotation>
+
+    <xsd:element name="cache" type="cache_type" />
```
wrygiel: If we're only trying to add attributes to an existing `groundspeak:geocache` element, then I guess we shouldn't declare our own `cache` element. Instead, we should declare only specific elements and annotate them properly (e.g. "this element can be put only as the child of the `groundspeak:geocache` element").

Similar is true for the `log` element below.

following5: Agree for the logs. They depend on groundspeak, because logs are added to GPX by the groundspeak extension.

`oc:cache` (as proposed) is independent from `groundspeak:cache`; it's directly attached to the `wpt` element, which represents a geocache (it has `<sym>Geocache</sym>`). There will probably be more Opencaching-specific cache attributes be added, so it may make sense to group them as `oc:cache`.

okapi/static/ocgpx/1/ocgpx.xsd
following5: I am fine with `https://github.com/opencaching/gpx-extension`.

(Had in mind that there was a case when the Github path changed for such a document and OKAPI needed an adjustment. But maybe I am confusing this with something else.)

okapi/static/ocgpx/1/ocgpx.xsd
wrygiel: Okay, sounds reasonable. `wpt` can also represent "additional waypoint", but we may use a different element for extending this one.

okapi/static/ocgpx/1/ocgpx.xsd
wrygiel: Even if GitHub paths change, they still redirect properly. In other words, the value of `xmlns` will always work properly. This is not necessarily true for the value of `xsi:schemaLocation`, such as [this one](https://github.com/erasmus-without-paper/ewp-specs-api-discovery/blob/v4.1.0/manifest-example.xml#L11), but these links *can* break, so it's not a big deal (in fact, we don't even need to provide `xsi:schemaLocation`, if we don't want to).

wrygiel:
See here:

https://github.com/opencaching/gpx-extension

okapi/static/ocgpx/1/ocgpx.xsd
wrygiel: BTW - worth considering - perhaps we could replace this be a more generic A-code attribute element, and introduce a new A-code for this purpose?

wrygiel: Since A-codes are quite solid and backward compatible, developers could implement this logic by checking if a particular A-code is present. Also, the users of application which already display attributes, would get this information independently displayed, without the need of the developer implementing support for the new A-code.

(I was quite sure we already had an A-code for "requires passoword", but I checked, and it seems we don't.)

following5: Some things to consider:

* There is already a way to deliver OC attributes in GPX, which is understood by some applications; it's called `attrs`=`gc_ocde:attrs` in OKAPI.
* The only major OKAPI App - c:geo - already processes the password requirement as indicated by the `req_passwd` flag of services/caches/geocache(s). Don't know how they display it.
* It may be a good idea to show the password requirement in a more prominent way than an attribute. Because if you are in the field and forget to write down the password, you will not be able to log the cache later. OCPL therefore prominently inserst a text into the GPX description text which points out the password reqirement. 
* It would be the first OKAPI-only attribute, because passwords are stored as strings, while OC attributes are stored as boolean flags. So it would need some change in the OKAPI attribute logic, and it could not be _set_ in the same way like other attributes (see the new [services/caches/edit](https://opencaching.pl/okapi/services/caches/edit.html) method).
* The user's requests that lead to this proposal was to transfer the "requires password" flag from www.opencaching.de via GPX to c:geo.

All things considered, I would prefer to go for the quick solution and replicate the `req_passwd` flag from services/caches/geocaches to GPX. Creating a new GPX attributes standard (besides of the already existing GC and OCDE standards) might be persued separately. Will do no harm if the "requires password" information would be available redundantly.

wrygiel: > Will do no harm if the "requires password" information would be available redundantly.

Yes, sure.

It's a pity I never found the time to finish the development of A-codes and S-codes. Probably never will... Despite this, my mind keeps finding new applications for them. :) If designed correctly, they would make such an elegant and forward-compatible model for expressing so many things...! ;)

Bananeweizen:
I'm not an expert in XML stuff, but I would leave away the GPX and Groundspeak namespace declarations from your extension, e.g. lines 4 and 5 of https://github.com/opencaching/gpx-extension/blob/master/schema.xsd#L4. You are not really using them in your schema, but you are fixing their version number by doing so. This is a problem because Groundspeak GPX files come in different versions, and therefore some of those would not fit your extension.

At least in theory, in practice basically no one actually verifies using the schemas from the namespace. :)

wrygiel:
I am referring to them, but only in the annotations, not the schema itself. You're right that they can be removed, but there's also nothing wrong with leaving them there, I think, and people who know XML Schema well, tend to look for prefix references there.

> This is a problem because Groundspeak GPX files come in different versions

Are you sure? If I remember correctly, they attempted to introduce a new version, but they have backed out of it quite quickly, because it turned out that there *were* some applications which were xmlns-dependant. (BTW, it is correct for the app to be xmlns-dependant, this is exactly why XML namespaces were created for, after all. But many developers don't understand XML namespaces.)

following5:
In the GC user profile, there is the choice of GPX version 1.0 and 1.0.1:

gcprofile

* http://www.groundspeak.com/cache/1/0/cache.xsd
* http://www.groundspeak.com/cache/1/0/1/cache.xsd

GSAK offers the option of GPX "1.0" (which actually procuced 1.0.1 files) and 1.1. If you select 1.1, it shows a warning:

gsak

OC sites only produce GPX 1.0.1.

following5:
http://www.topografix.com/GPX/1/1/gpx.xsd

following5:
The version 1.1 GSAK refers to is _Topografix_ 1.1: http://www.topografix.com/GPX/1/1/gpx.xsd

wrygiel:
If I understood you correctly, you're saying that:

* GC still allows users to download GPX with groundspeak's 1.0 schema (but I guess that the default option is to use the 1.0.1 one, right?)
* GSAK strongly suggests groundspeak's 1.0.1 schema, doesn't allow 1.0 schema, and warns about using 1.1 schema.
* Not sure what you means by "refers to", when you're saying that GSAK refers to is Topografix 1.1? Does  it allow to generate GPX files in Topografix 1.1 format?

I don't gave a GC account so I might be missing something, but I was quite sure that "Topografix 1.0" + "Groundspeak 1.0.1" was the only true geocaching GPX standard, for many years now. And that the attempt to introduce "Groundspeak 1.1" was a failure.

following5:
Sorry, I confused Topografix and Groundspeak versions. To sort things:

* GC produces Topografix 1.0 GPX with either Groundspeak 1.0 oder Groundspeak 1.0.1 extension. Yes, the default setting for new GC accounts is 1.0.1, I guess for about 6 or 7 years now.

* GSAK produces either Topografix 1.0 or 1.1 GPX, both with Groundspeak 1.0.1 extension, and strongly recommends Topografix 1.0.

> I was quite sure that "Topografix 1.0" + "Groundspeak 1.0.1" was the only true geocaching GPX standard

Surely it is the prevailing standard today, while some GC users will still download Topografix 1.0 + Groundspeak 1.0 (which doesn't know about geocache attributes). I don't see any newer version of Groundspeak extension out there.

wrygiel:
This deserves an article :)

https://github.com/opencaching/gpx-extension/blob/master/all-these-namespaces.md

Bananeweizen:
> BTW, it is correct for the app to be xmlns-dependant, this is exactly why XML namespaces were created for, after all. But many developers don't understand XML namespaces, hence the problems

I wish everyone would understand that, and not blame cgeo for not importing GPX files, which are - from an XML point of view - not really GPX files. We meanwhile filter the underlying byte stream to get rid of some nasty things that never ever should appear in XML files. :)

okapi/services/caches/formatters/gpx/gpxfile.tpl.php
```diff
@@ -15,7 +15,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns:groundspeak="http://www.groundspeak.com/cache/1/0/1"
     xmlns:ox="http://www.opencaching.com/xmlschemas/opencaching/1/0"
-    xmlns:oc="https://opencaching.pl/okapi/static/ocgpx/1"
+    xmlns:oc="https://raw.githubusercontent.com/opencaching/gpx-extension/stable-v1"
```
wrygiel: Bad xmlns

okapi/services/caches/formatters/gpx/gpxfile.tpl.php
```diff
@@ -27,8 +27,8 @@
         http://www.opencaching.com/xmlschemas/opencaching/1/0
         https://raw.githubusercontent.com/opencaching/okapi/master/etc/nsox.xsd
 
-        https://opencaching.pl/okapi/static/ocgpx/1
-        https://opencaching.pl/okapi/static/ocgpx/1/ocgpx.xsd
+        https://raw.githubusercontent.com/opencaching/gpx-extension/stable-v1
```
wrygiel: Here too.

okapi/services/caches/formatters/gpx/gpxfile.tpl.php
wrygiel: https://github.com/opencaching/gpx-extension/issues/6

teiling88:
> Would you support this OC GPX extension on your websites, i.e. in native GPX generation in OCPL / OCDE code? I may also do the implementations for that.

do it :-)

andrixnet:
OCPL also has support for log password.

andrixnet:
Furthermore, if OC specific information is considered to be included as having own namespace in GPX files, then I would recommend the following as well: 

OC has different / more cache types. One could declare such cache types using the namespaces according to `Groundspeak 1.0.1` which is the most compatible, also in doing so, allows showing in apps that do not know about OC cache types a compatible equivalent, and define a separate OC namespace with the OC cache type. This should be treated by clients as: if present and understood by the app, OC namespace definition to replace the generic definitions. 

Like: 
```
<type>Geocache|Unknown Cache</type>
<groundspeak:type>Unknown Cache</groundspeak:type>
<oc:type>Moving Cache</oc:type>
```

Non-OC aware clients will show "unknown cache".
OC aware clients will show "moving cache".

Same goes for cache sizes, log types, waypoint types and especially attributes. 
OC values that can be equivaleted have both namespaces in exported GPX, OC values that have no equivalent (for example attribute OCUK's "rated on handicaching.com") be shown only by OC aware apps.

following5:
> OCPL also has support for log password.

OCPL inserts a localized text into the description text:

"WARNING! You need a PASSWORD to log this cache. You have to either find in the cache (and remember to write it down) or obtain it according to the description!!!"

This is something different than a "requires password" field in the GPX. The latter one will e.g. allow c:geo to prompt for a password then logging the cache. Currently you cannot log "found" GPX-imported caches at all with c:geo if they require a password.

andrixnet:
Yet the presence of this field is one step towards making that happen, right?

SammysHP:
The problem is that some users use tools to manage their caches like GSAK. The additional information will be lost as long as the tools do not work with the extended schema.

following5:
@SammysHP 
What method would you prefer to transfer the information "requires password" from an OC site through GPX to c:geo?

SammysHP:
With the proposed GPX extension and a request to other tools. ;)

wrygiel:
> The problem is that some users use tools to manage their caches like GSAK. The additional information will be lost as long as the tools do not work with the extended schema.

In general, external tools should avoid editing/removing XML elements they don't understand. Perhaps that's something that can be reported to GSAK authors as a feature-request.

SammysHP:
All tools that I know store the data in some sort of internal data structure that has a fixed set of fields. Some tools that should be notified: GSAK, Cachewolf, c:geo (already done), QLandkarte GT, QMapShack, Locus. I don't know any other tool that works explicitly with OpenCaching.

wrygiel:
> All tools that I know store the data in some sort of internal data structure that has a fixed set of fields.

I think so too. Because this is the easiest approach. It can be done in a better way, but it requires some work (in particular, the internal data structure would need to be able to hold "unknown XML elements"). And most authors probably won't be willing to do this work, but still, perhaps some good ones might :)

PR #517, opened 2017-10-05
PSR-2 proposal

teiling88:
A proposal for the integration of PSR-2 Standard with php-cs-fixer.


following5:
This does not comply to the [OKAPI style guide](https://github.com/opencaching/okapi/blob/master/etc/CODESTYLE.md).

teiling88:
Yes that's right. But I talked with @wrygiel about a more Standard Style and created a proposal for that. (I can not find the link unfortunately). So I think we should talk about this pr instead of closing it.

following5:
I think it would be easier to discuss the Okapi style guides than a change of 3600 code lines. Find out which Okapi rules conflict with PSR-2, and decide which of them we want to keep and which to discard.

But ok, let's try it here.

okapi/core/Cache.php
```diff
@@ -129,10 +123,13 @@ public static function get_many($keys)
...
-        if (count($dict) < count($keys))
-            foreach ($keys as $key)
-                if (!isset($dict[$key]))
+        if (count($dict) < count($keys)) {
+            foreach ($keys as $key) {
+                if (!isset($dict[$key])) {
                     $dict[$key] = null;
+                }
+            }
+        }
```
following5: Lots of boilerplate code introduced here - seven code lines instead of three, without any improvement to readability.

okapi/core/Okapi.php
```diff
@@ -1030,21 +1022,47 @@ public static function logtypeid2name($id)
...
+        }
+        if ($id == 11) {
+            return "Temporarily unavailable";
+        }
+        if ($id == 12) {
+            return "OC Team comment";
+        }
+        if ($id == 13 || $id == 14) {
+            return "Locked";
+        }
```
following5: Huge code inflation here. The original code much more readable.

okapi/services/caches/formatters/ggz/ggzindex.tpl.php
```diff
@@ -9,14 +9,15 @@
...
+    ?>
         <file>
             <name><?= $f['name'] ?></name>
             <crc><?= $f['crc32'] ?></crc>
             <time><?= date('c') ?></time>
             <?php
                 foreach ($f['caches'] as $c) {
-            ?><gch>
+                    ?><gch>
                 <code><?= $c['code'] ?></code>
```
following5: This looks strange, as well as the change 8 lines above. I guess the reformatting tool cannot deal well with these templates.

okapi/core/Cache.php
following5: My comment is not undestandible without the previous lines. Suggest to open a separate issue for discussion.

okapi/core/Okapi.php
following5: My comment is not undestandible without the previous lines. Suggest to open a separate issue for discussion.

okapi/core/CronJob/CheckCronTab2.php
```diff
@@ -16,46 +16,48 @@ public function get_period()
...
 
-        # There was no ping. Decrement the counter. When reached zero, alert.
+        // There was no ping. Decrement the counter. When reached zero, alert.
 
         $counter = Cache::get('crontab_check_counter');
         if ($counter === null) {
             $counter = 5;
         }
-        $counter--;
+        --$counter;
```
following5: 👍

okapi/core/CronJob/CronJob.php
```diff
@@ -28,7 +28,7 @@ public function get_name()
      * ONLY ONCE per cronjob execution, directly AFTER the job was run.
      * (Scheduling the *first* run of `execute` method is out of this method's
      * control, but you can prevent the actual job from being executed by using
-     * custom conditions in your `execute` method.)
+     * custom conditions in your `execute` method.).
```
following5: -.-

okapi/core/CronJob/Cron5Job.php
```diff
@@ -25,6 +25,7 @@
     public function get_next_scheduled_run($previously_scheduled_run)
     {
         $t = time() + $this->get_period();
-        return ($t - ($t % 300));
+
+        return $t - ($t % 300);
     }
```
following5: I agree in this case, but generally I would prefer to allow some bracketing of assigned terms. E.g. for terms containing `==` and for `? :`.

okapi/core/CronJob/CacheCleanupCronJob.php
```diff
@@ -49,27 +50,27 @@ public function execute()
...
+        // but never read will be removed after 2 days (0.08^2 < 0.01 < 0.08^1).
 
         Db::execute("
             delete from okapi_cache
             where
                 score is not null
                 and score < '".Db::escape_string($limit)."'
         ");
-        Db::query("optimize table okapi_cache");
+        Db::query('optimize table okapi_cache');
```
following5: No. SQLs should be consistently quoted, not some in double quotes and some in single quotes.

okapi/lib/ClsTbsZip.php
```diff
@@ -15,13 +15,12 @@
...
 
-
 class ClsTbsZip
 {
     const TBSZIP_DOWNLOAD = 1;   // download (default)
     const TBSZIP_NOHEADER = 4;   // option to use with DOWNLOAD: no header is sent
-    const TBSZIP_FILE     = 8;   // output to file  , or add from file
-    const TBSZIP_STRING   = 32;  // output to string, or add from string
+    const TBSZIP_FILE = 8;   // output to file  , or add from file
+    const TBSZIP_STRING = 32;  // output to string, or add from string
```
following5: In this small example I think it doesn't matter, but in larger code sections a tabular formatting can improve readability.

following5: See for example [these new tables](https://github.com/opencaching/okapi/blob/master/okapi/services/caches/formatters/gpx/WebService.php#L27).

teiling88:
I love standards so my main question is, https://github.com/opencaching/okapi/pull/517/commits/b313a0f14b78f2fad8f2569cd0b2bc47c61cdc96 does it feel good for okapi developers to have PSR-2 Standard? The other commits are only proposals from my side.

okapi/lib/ClsTbsZip.php
teiling88: > a tabular formatting can improve readability.

but a small change can be exploded to many lines if we add as an example const TBSZIP_STRING_SUPER_LONG = 1024; every other line has to be updated too.

following5: I don't see it that strict. In many cases tabular formatting is nice, in some cases it won't work at all, and in other they may be a few strings that don't fit but the rest of the table is aligned. It's an _option_.

following5:
I think that standards a a good thing, but if there is a way to improve them, then it's fine to create own standards. There would be no progress and no innovation if people would always comply to existing standards.

I feel ok with PSR-2 with three exceptions where it produces less readable or boilerplate code.

1. Control blocks _must_ be bracketed. I would prefer to keep the Okapi style rule that brackets can be omitted for simple control structures:

```
foreach ($basic_fields as $field)
    if (!in_array($field, $fields_array))
        foreach ($results as &$result_ref)
            unset($result_ref[$field]);
```

PSR-2:

```
foreach ($basic_fields as $field) {
    if (!in_array($field, $fields_array)) {
        foreach ($results as &$result_ref) {
            unset($result_ref[$field]);
        }
    }
}
```

2. Opening bracket _must_ go on the same line.  I would prefer to keep the Okapi style rule and allow placing it on the next for large blocks (or for improved readability).

3. This is the most important one. You did not apply it in b313a0f. 100% PSR-2 would mean that multiline function parameters _must_ start on the next line. This negatively impacts SQLs.

Okapi style:

```
    Db::execute("
        delete from okapi_nonces
        where
            timestamp < unix_timestamp(date_add(now(), interval -6 minute))
            or timestamp > unix_timestamp(date_add(now(), interval 6 minute))
    ");
```

PSR-2:

```
    Db::execute(
        "
        delete from okapi_nonces
        where
            timestamp < unix_timestamp(date_add(now(), interval -6 minute))
            or timestamp > unix_timestamp(date_add(now(), interval 6 minute))"
    );
```

Alternative PSR-2:

```
    Db::execute(
        "delete from okapi_nonces
         where
            timestamp < unix_timestamp(date_add(now(), interval -6 minute))
            or timestamp > unix_timestamp(date_add(now(), interval 6 minute))"
    );
```

The OKAPI SQLs are the cleanest and easiest I have ever seen. Would not want to give that up.

teiling88:
so nobody else seems to be interested in it so i will close it.

PR #524, opened 2017-11-21
Update Okapi.php

Rotzbua:
changed to strict compare to avoid misbehavior, if hash may changed in future

Example:
```php
var_dump(md5('240610708')=='0e46');
bool(true)

### DIFF
--- a/okapi/core/Okapi.php
+++ b/okapi/core/Okapi.php
@@ -1162,7 +1162,7 @@ class Okapi
     public static function require_developer_cookie() {
         if (
             (!isset($_COOKIE['okapi_devel_key']))
-            || (md5($_COOKIE['okapi_devel_key']) != '5753f318c1495c01637f7f6b7fc9c5db')
+            || (md5($_COOKIE['okapi_devel_key']) !== '5753f318c1495c01637f7f6b7fc9c5db')
         ) {
             header("Content-Type: text/plain; charset=utf-8");
             print "I need a cookie!";

### CODE REVIEW

**Comment by following5** [comment]
The [Streisand effect](https://en.wikipedia.org/wiki/Streisand_effect) of merging this would be higher than the benefit of fixing this hypothetical problem.

**Comment by following5** [comment]
(But we may include the fix later in another commit.)

**Comment by Rotzbua** [comment]
@following5 Actually there is no bug or security issue. There could be one if some developer **change** the hash and the new hash starts with `0e`. 

This is fine:
```php
(md5($_COOKIE['okapi_devel_key']) !== '5753f318c1495c01637f7f6b7fc9c5db')
```

That would be a problem:
```php
(md5($_COOKIE['okapi_devel_key']) !== '0e462097431906509019562988736854')
```

Because of week compare, php would interpret the right side as an integer 0<sub>462097431906509019562988736854</sub> which is in fact `0`. Then the left side could be manipulated so that there is also a hash starting with `0eXXXXX` so it would be also interpreted as `0`. A compare of two `0` would result as a `true`.

So you can merge my patch, because there is actually no security issue and it would be good to have an Stresand effect so people would prefer strict compares in php ;) .

--
Sorry for the missing source:
https://www.cryptologie.net/article/268/how-to-compare-password-hashes-in-php/

Rotzbua:
@following5 Any update?

following5:
Have included this change with some other commit, for the reason stated above. So this hypothetical problem is fixed now.

PR #539, opened 2018-09-25
Update German language files

harrieklomp:
Cleaned up version German version of LC_MESSAGES.
From here all other languages can be updated.


following5:
I have cleand up also the other translation files.

Issue #542, opened 2018-10-12
SearchAssistant.php performance at mariaDB


kojoty:
@following5 @andrixnet @harrieklomp @deg-pl 

Last days OCPL has moved to new server (provider) and as a side change we switch from mysql to mariaBD. After the change performance of the map was dramatically reduced - tails were generated with huge cpu load and delay.

I've spent some time and that's the results:

1. it seems that the most expensive query for map tiles generation is here:
https://github.com/opencaching/okapi/blob/master/okapi/services/caches/search/save/WebService.php#L135

It is called only once per search query (or map filter setting) but it takes often 10 or more seconds to get the results what blocks map tile generation.

2. we started to look in details how it works and it seems that the problem is in very long list in IN condition - query like:
`select ... from ... where cache_id <not> in ( <and-very-long-list-of-ids-here> )`

This long list is generated by separate query which gets for example all caches found by user etc.

For OCPL (with 60k+ geocaches) often this list can have for example almost all caches - but DB has problem with list longer than thousand of positions.

3. ...but (what is ￼surprising for me - I'm not DB expert) same query but with same subquery in place of very-long-list-of-values works like a charm - time of response is many times shorter - often < 1s.

4. I tested it at production (at OCPL) and it seems that map works much better after small fixes in SearchAsistant.php class

5. if you want know more details I can share - it seems that there is significant difference between mysql and mariaDB in the way of handling such type of queries - at least in my local configurations SQL EXPLAIN command shows that both db engines handle it differently.

**Conclusions:**
We would like to apply changes to OKAPI switch SearchAsistant code to get "subqueries" instead of "very-long-lists" to but I'm not sure what impact it can have for other nodes.

It seems to be great improvement for OCPL but for other nodes it can reduce performance - I'm not sure all details of configuration and tuning magic made by admins of other nodes.

**How to deal with this?**

I'll try to deliver PR with proposition of changes soon.

kojoty:
code change proposition delivered in PR above...

following5:
With MySQL and older MariaDB, subqueries were very inefficient. I guess that's why Wrygiel implemented it this way. With newer MariaDB versions, subqueries are working fine.

All OCDE systems including developer VMs are running new MariDB, so it should be ok. I will do tests on my develsite.

OCDE doesn't use the OKAPI tile map, maybe that's the reason that the performance problem was not detected there.

kojoty:
[...]

PR #543, opened 2018-10-12
#542: proposition of sub-query use in SearchAsistant.php

kojoty:
This is proposition of change - details here: #542


following5:
Sorry for the delay. The change is fine, and on my OCDE develseite I did not detect any performance difference between both settings.

Issue #523, opened 2017-10-22
Retrieve log entry creation date


following5:
As log entries can be back-dated, the log entry date does not always reflect the date when the log was submitted. The old OC "XML interface" already exposes the date when log entries were submitted, and it is also exposed through the sorting in the "new logs" lists of OC sites. So this is no private information. Just had a case where I needed to know the submission date of a log to analyze a bug. I guess there are some other use cases for this.

Time should not be exposed for privacy reasons.

following5:
Same applies to the log's last modification date.

PR #545, opened 2018-10-27
add log entry fields 'date_created' and 'last_modified'; updates #523

following5:
This enables Okapi developers to retrieve the information 'date_created' and 'last_modified' for log entries.

Would you prefer to expose the complete date and time here through Okapi, or just the date? OCDE prefers date-only for privacy reasons.


mzylowski:
I think date-only is pretty enough, so:
+1 to date-only for OCPL

following5:
Okay. Then we can keep it simple and define the new fields as date-only.

PR #553, opened 2018-11-05
OCPL log text InputFilter class OKAPI usage refactor

kojoty:
@following5 

This is small fix for inputFilter usage.
Please note I don't have OCDE dev setup to test if it works there - but I hope there shouldn't be any issue..

PR #554, opened 2018-11-06
OCPL log text InputFilter class OKAPI usage refactor

kojoty:
@following5 

what do you think about such solution?
If it is acceptable I need to deliver okapi_settings to OCPL code before merge of it.

BTW - I found and fixed small bug - _call should return array...

(as continuation of #553)

### DIFF
--- a/okapi/Facade.php
+++ b/okapi/Facade.php
@@ -42,21 +42,19 @@ use okapi\lib\OCSession;
 
 require_once __DIR__ . '/autoload.php';
 
+OkapiErrorHandler::init();
+Okapi::init_internals();
+
+$extAutoloader = Settings::get('EXTERNAL_AUTOLOADER');
+if ($extAutoloader) {
+    require_once $extAutoloader;
+}
+
 /**
  * Use this class to access OKAPI's services from external code (i.e. OC code).
  */
 class Facade
 {
-    static $initialized = false;
-    private function init()
-    {
-        if (!self::$initialized) {
-            OkapiErrorHandler::init();
-            Okapi::init_internals();
-            self::$initialized = true;
-        }
-    }
-
     /**
      * Perform OKAPI service call, signed by internal 'facade' consumer key, and return the result
      * (this will be PHP object or OkapiHttpResponse, depending on the method). Use this method
@@ -65,7 +63,6 @@ class Facade
      */
     public static function service_call($service_name, $user_id_or_null, $parameters)
     {
-        self::init();
         $request = new OkapiInternalRequest(
             new OkapiFacadeConsumer(),
             ($user_id_or_null !== null) ? new OkapiFacadeAccessToken($user_id_or_null) : null,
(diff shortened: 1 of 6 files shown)


following5:
Okay! Have added settings validation.

following5:
Oops, there is an issue with `Settings::get()`. When called via Facade that early, this will follow:
* Settings::load_settings()
* OkapiErrorHandler::disable()
* restore_error_handler()

... without a preceding `set_error_handler()`. This will drop the OC errorhandler, if one was installed (OCDE does).

What about this:

```
$ocAutoloader = __DIR__ . '/../vendor/autoload.php';
if (file_exists($ocAutoloader))
    require_once $ocAutoloader ;
```

following5:
Generally: Okapi error handling should be enabled before any Okapi functions are called (like Settings:get), but it should not be enabled before any Facade function is called. This is difficult to achieve if Settings:get is called from autoload.php.

following5:
Generally,  it's a hen-and-egg problem between 1. initializing the autoloader, 2. inizializing Okapi errorhandling and 3. calling Settings::get(). No 2 should be done before no 3, but needs no 1, which does no. 3. :)

following5:
Have merged the new log editing methods, which conflich with this commit. Log text processing has moved to LogsCommon.php. So this change must be re-applied.

kojoty:
sure :) 
I have a busy days last days and this change has a low priority - I'll take care of it.

kojoty:
OK, I updated this PR. To make it works at OCPL code we need to first add ClassPathDictionary.php to okapi_settings.php

following5:
It still calls `Settings::get()` prior to `OkapiErrorHandler::init()`, which can cause trouble.

What about my suggestion to hard-code it in the same way as [this](https://github.com/opencaching/okapi/blob/master/okapi/Settings.php#L254)?
```
$ocAutoloader = __DIR__ . '/../vendor/autoload.php';
if (file_exists($ocAutoloader))
    require_once $ocAutoloader;
```

following5:
I have tested your recent commit with this setting ...
```
'EXTERNAL_AUTOLOADER'=> __DIR__.'/vendor/autoload.php'
```
and also tested direct inclusion as proposed above. Still the same error: "Class 'Utils\Text\UserInputFilter' not found" :( Did I miss something?

kojoty:
> I have tested your recent commit with this setting ...
> 
> ```
> 'EXTERNAL_AUTOLOADER'=> __DIR__.'/vendor/autoload.php'
> ```
> and also tested direct inclusion as proposed above. Still the same error: "Class 'Utils\Text\UserInputFilter' not found" :( Did I miss something?

Nope, for OCPL for now it should be: 
'EXTERNAL_AUTOLOADER'=> __DIR__.'/lib/ClassPathDictionary.php'

There are still a few "legacy" classes - after refactoring of this code OCPL will use composer autoloader, but for now this is necessary.

following5:
Okay! Have added a commit which fixes the errorhandler initialization thing.

Issue #398, opened 2016-03-30
Query OC site parameters through OKAPI


following5:
This is a spin-off from [this issue](https://github.com/opencaching/opencaching-pl/issues/412) in OCPL project.

SammysHP wrote:

> if platforms decide to provide different subsets of the OKAPI functionality, this must be declared in the API interface and signaled to c:geo in some way.

I answered:

> The OKAPI philisophy is to compensate for differences between OC sites as far as possible, so that client developers need not to care about that. 
> ...
> But yes, there are some other differences between OC sites which OKAPI is not able to compensate for, so I agree that the feature of signaling site capabilites to the clients should be added to OKAPI. 

Wrygiel replied:

> I'm afraid adding this would be both very difficult and backward-incompatible.

There are lots of differences in the behaviour of OC sites. OKAPI currently does not explain which site behaves like what, so client software cannot be adjusted to that. E.g. if a was an application developer, I would like to disable or hide the `rating` and `powertrail` search options when searching on an OC site that do not implement these features.

My idea is to create a "site attributes" dictionary - an XML file in the 'etc' directory, which may effectively contain values like these for OCDE

```
service.caches.search.rating = unavailable
service.caches.search.powertrails = unavailable
service.caches.search.cache_types = Traditional,Multi,Quiz,Math/Physics,Virtual,Event,Webcam,Moving,Drive-In,Other
service.logs.submit.needs-maintenance-2logs = false
service.logs.submit.type.plaintext = functional
```

and these for OCPL:

```
service.caches.search.rating = available
service.caches.search.powertrails = available
service.caches.search.cache_types = Traditional,Multi,Quiz,Virtual,Event,Webcam,Moving,Podcast,Other
service.logs.submit.needs-maintenance-2logs = true
service.logs.submit.type.plaintext = dysfunctional
```

The method's docs would document which site attributes can be queried to find out the behaviour for the current site, through a method like 'apisrv/site_parameter'.

Parameters should only be exposed for things where developers will substancially benefit from it, e.g. it should not be exposed if a site provides preview images or alt_wpts.user-coords. Except if OKAPI implements the feature "submit caches", then this would be useful.

wrygiel:
> if a was an application developer, I would like to disable or hide the rating and powertrail search options when searching on an OC site that do not implement these features.

Ok, perhaps you're right and some of these attributes would make sense. But we should not require developers to read them (this should be optional task, only for the most pedantic ones).

> an XML file in the 'etc' directory

[...]

Issue #220, opened 2015-07-19
document all known cache types


GoogleCodeExporter:
```
(issue 210)
> If we decide to expose something via OKAPI, then I think we should try to 
document it as well as we can.

So the other cache types should be documented, at least those which will not be 
discarded in the near future:

* Moving cache (PL, DE)
* Webcam cache (PL, DE)
* Podcast cache (PL)
* Drive-In cache (DE)

```

Original issue reported on code.google.com by `following09` on 14 Apr 2013 at 4:30

following5:
"Unknown" type is also missing in the OKAPI docs, though OKAPI itself returns it for unknown types.

following5:
Done:
* added documentation of Moving and Webcam cache types
* allow to search for types not available at the OC installation, and ignore them

Types to do:
* Other
* Drive-In, Math/Physics (OCDE only)
* Podcast, Own (OCPL only)

following5:
fixed by 3b2ebae1

Issue #440, opened 2016-10-04
Update/ Create Listing via OKAPI


protux:
Hi,

for some apps it would be useful to update a cache-listing via okapi. e.g. one problem are double listings which are published on OC and GC but will only be maintained on GC. An app could automatically offer to keep an eye on the GC listing and update the OC listing.

Without the api one would need to store the users credentials what would not be good. So this "quality assurance" tool could only exist with the help of the OKAPI.

Another use case could be if a user is fed up with GC and whants to move his account to OC (or maybe he decides to create double listings of all his caches). If he wants to move his caches he would need to do that manually, with OKAPI support one could automate that.

What do you think about listing manipulation via OKAPI?

wrygiel:
Adding new log entries is already implemented. So, what you propose is to add methods for editing past log entries, right?

wrygiel:
I feel there are too many issues in one issue here. :) I got confused a bit.

protux:
No, my proposal aims more for editing the cache description incl. D/T Value, coordinates, Spoilerimages etc. :) Maybe listing was the wrong word, excuse me, please.

wrygiel:
I'm afraid this won't be possible at this point. There are too many differences between both OC branches to allow us to create one consistent API for that.

And even if this changed, it would still be a huge task and, sadly, we probably won't have resources to ever do this. (I even think there already was an issue for that.)

following5:
#488 is a subissue of this, adding a method to change geocache listings via OKAPI. It starts with changing passwords, which is rather simple.

Expanding that to editing or submitting whole geocache listings will be very expensive, as Wrygiel wrote. But I think it's possible, the (relative) differences between branches are similar to submitting logs.

And this feature is really needed, exactly for the reasons that protux gave. For OCDE there has been a tool that copies GC listings to OC by directly interacting with the OC.de website, but it has recently been broken by OCDE code changes and probably won't come back. OKAPI would be the logical way to implement a replacement.

following5:
Editable properties of a Geocache:
* name (only one name on OC sites, names in multiple languages in OKAPI)
* type
* container size
* coordinates
* country
* OCPL: region (automatically calculated in OCDE)
* difficulty
* terrain
* trip time
* trip distance
* GC code
* OCPL: NC code
* OCPL: TC code
* OCPL: GPSGames code
* attributes
* descriptions in multiple languages
* short descriptions in multiple languages
* hints in multiple languages
* images
  * title
  * binary image data
  * position
[...]

PR #561, opened 2018-11-10
Proposal for services/caches/capabilities and overhaul of cache type specs

following5:
This change addresses issues #220 and #398. It builds on commit 
9b46a60.

* Add the fields `geocache_types`, `geocache_sizes` and `geocache_ratings` to `apisrv/installations`. This enables developers to dynamically adjust their user interface to OC site capabilities, especially when searching for caches.
* Give a clearer and firmer interface to available geocache types.

This proposal deviates from two original OKAPI paradigms: "OC sites have too many cache types by far", and "keep differences between OC branches out of the interface, if possible". So here is a request for comments.

### DIFF
--- a/okapi/core/Db.php
+++ b/okapi/core/Db.php
@@ -299,6 +299,8 @@ class Db
                 table_schema='".self::escape_string(Settings::get('DB_NAME'))."'
                 and table_name='".self::escape_string($table)."'
                 and column_name='".self::escape_string($field)."'
-        ");
+        ") + 0;
+        /* OC.DE installation returned a string here, OC.PL an integer.
+         * Added type cast. */
     }
 }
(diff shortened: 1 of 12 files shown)


following5:
Have refactored this proposal to a new method services/caches/capabilities, similar to services/logs/capabilities (#562). This gives "room" and flexibility for adding more fields. #440 (submitting and editing caches) would need:
* list of available cache status
* list of available countries
* list of available regions
* list of available waypoint types
* avaliable image and waypoint properties

and some flags on installation-dependend fields.

Also moved the max. password length field from apisrv/installation to the new method. it had wrong data type (string) at OCDE, and so far was not used at OCPL sites. So we can safely change the interface here.

kojoty:
@following5 what exactly you mean by: "list of available countries,  list of available regions"?

following5:
This ...

editcache

and this:

searchcache

If we add those features to OKAPI - search by country/region, or submit and edit caches - there must be some list of countries and regions that the user can choose from.

kojoty:
OK, cool - so it will allow to add in future search by country/region through OKAPI, am I right?

following5:
It would allow that, yes. I have not implemented the country and region lists here, just considered what we may need to add in the future.

Issue #552, opened 2018-11-05
Found and Attended logs must update OCPL Merit Badges


following5:
In March 2017, the optional "Merit Bages" feature was added to OCPL code; see 
[MeritBadgeController.php](https://github.com/opencaching/opencaching-pl/blob/master/lib/Controllers/MeritBadgeController.php). It is enabled e.g. at Opencaching.PL.

`services/logs/submit` must update the user's badges for new "Found it" and "Attended" logs. Same with editing log types and deleting logs; see #421.

Funny that noone noticed the data lack for 18 months. Maybe because OCPL native code recalculates the bage data after every change, so it is fixed with the next on-site Found/Attended log.

mzylowski:
`Funny that noone noticed the data lack for 18 months`
@tripper1971 Noticed, but ignored this :)

kojoty:
@following5 are you sure that there was a "data lack" - I believe (not sure) that everything is counted well, only user don't get notification about the new badges on time.

When user submit log next time (but directly by website - not by OKAPI) the notification appears.

We had known about this issue but - yes - ignored it because it needed changes at OKAPI side and in fact was a minor (corner case) thing.

following5:
I am sure that the `badge_user` table is not updated by OKAPI. Will be correctly updated when the user submits, edits or deletes another "Found it" or "Attended" log on the website.

So if a user only logs by an Okapi App, I think there will be no merits at all ...?

kojoty:
> So if a user only logs by an Okapi App, I think there will be no merits at all ...?
If "only" by OKAPI app - I agree - you're right. But after the first log by websites all badges will be applied.
Creating a log is only a trigger for recalculating the stats.

But generally - I agree - we should prepare some solution for it.

mzylowski:
> When user submit log next time (but directly by website - not by OKAPI) the notification appears.

> So if a user only logs by an Okapi App, I think there will be no merits at all ...?

Almost all my logs are added with OKAPI. When I added one log via the website a few weeks ago https://github.com/opencaching/opencaching-pl/issues/1616 happened
So both issues are related.

following5:
b9a86651 adds a signalling mechanism, which hands merit badge update requests over to OCPL code. There they can be processed via cronjob.

PR #570, opened 2018-11-22
Edit cache listings

following5:
This is work in progress for an update of `services/caches/edit`. As it is a BIG update, I put it here for the case there are comments / suggestions on the interface or implementation.

* Includes new options/fields for `services/caches/edit` , `services/caches/capabilities` and `services/attrs/attribute`.
* Todo: Add an option to `services/caches/geocache` to exclude the OCDE "listing is outdated" notice from the description - important for editing descriptions.
* Todo: Add edit options for country, region, additional waypoints and images.

### DIFF
--- a/okapi/Settings.php
+++ b/okapi/Settings.php
@@ -52,6 +52,12 @@ final class Settings
         'SITELANG' => "en",
 
         /**
+         * Some sites have even a list of language preferences, e.g. to
+         * initialize the language selection list when creating a new cache.
+         */
+        'SITELANGS' => null,
+
+        /**
          * If set, it will be passed to date_default_timezone_set.
          *
          * If your system is configured properly, then you are not required to
@@ -329,6 +335,8 @@ final class Settings
             $dict['SITE_LOGO'] = $dict['SITE_URL'] . 'okapi/static/oc_logo.png';
         if ($dict['JPEG_QUALITY'] < 50 || $dict['JPEG_QUALITY'] > 100)
             throw new Exception('JPEG_QUALITY must be between 50 and 100.');
+        if ($dict['SITELANGS'] === null)
+            $dict['SITELANGS'] = array($dict['SITELANG']);
 
         # The OKAPI code is only compatible with utf8 and utf8mb4 charsets.
         if (!in_array($dict['DB_CHARSET'], array('utf8', 'utf8mb4'))) {
--- a/okapi/core/Okapi.php
+++ b/okapi/core/Okapi.php
@@ -15,6 +15,7 @@ use okapi\core\Request\OkapiInternalRequest;
 use okapi\core\Request\OkapiRequest;
 use okapi\core\Response\OkapiHttpResponse;
 use okapi\Settings;
+use Utils\Text\UserInputFilter; // OCPL specific
 
 /** Container for various OKAPI functions. */
 class Okapi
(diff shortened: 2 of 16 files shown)


kojoty:
@following5 

I wonder if cache editing by OKAPI is a good idea for us. 

Generally this is great move, but how to be sure (in generally) that all operations are done when user edits cache by OKAPI or by webpage. 

This duplication of access ways is confusing for me. I think it would be good to think together how to do that... If we want to allow for edit we need to add more links between OKAPI and OCPL code...

Example of the problem:
- right now we have something like "altitude" for every cache.
- this is retrived from external service after every edit of cache coords (in OCPL code)
- how to deal with it in OKAPI? Does OKAPI code will update altitude as well or after OKAPI update altitude will be wrong?

following5:
> Generally this is great move, but how to be sure (in generally) that all operations are done when user edits cache by OKAPI or by webpage

That's a general problem with OKAPI: We always need to make sure that it works the same as on the native OC site. Redundancy _may_ be avoided by consolidating logic into OKAPI and calling it via Facade. If OCPL developers are interested, I could help with that.

Regarding altitude: We could add that to the [NPA cronjob](https://github.com/opencaching/opencaching-pl/blob/master/util.sec/cron/modules/cache_npa_areas.class.php). When cache coordinates are changed, a DB trigger sets `caches.need_npa_recalc = 1`, and then this cronjob calculates the NPAs. Could also call the altutide calculation function.

As long as that does not work, we could restrict coordinate editing to OCDE.

following5:
The rational for OKAPI cache editing is to allow easy on-site geocache maintenance for the owners. E.g. loggers complain about off-coordinates => owner goes to the cache, opens the cache in mobile app and clicks "correct coordinates". New coordinates are measured and automatically stored. Could do the same with additional waypoints. Could even record the stages of a new multicache on-site, directly adding them to the prepared listing.

following5:
> it can be complete different that OCDE approach - what complicates OKAPI code much...

Actually, there are more differences in logging between OCDE/OCPL than in cache listings. Implementing the listing editor is simpler than logging.

> But if we starting to migrate more and more "logic" to the OKAPI 

That's not necessary at all for cache editing. Was just a general - and somewhat offtopic - suggestion to avoid redundancy. For the "altitude", my suggestion is to do it other way round an have the logic only in OCPL code. (Btw, for the "Merit Badge" issue #552, I would prefer the same.)

following5:
I have editit my post above and striked-though the thing that you commented. It was offtopic.

following5:
But I fully understand that you feel uncomfortable with creating more logic dependencies between OKAPI and OCPL. My goal was to make OKAPI complete -- I guess it's 80% complete now, and I would need 2-3 more months to make it 100%. But I can't do that without your support.

kojoty:
@following5, I just share my minds...

like I told before I really appreciate your afford in OKAPI improvement.
If you need my support in any thing I will do what I can.  
I trust your perception in context of editing - let's go further with it.

kojoty:
side thought: 

> owner goes to the cache, opens the cache in mobile app and clicks "correct coordinates".

I little afraid about accuracy of such measurements. I usually uses apps like gpsAverage to read coords by smartphone and it often gets results far from expectation... 
...but probably developers of the client apps should worry about it - not we :)

following5:
Hmm okay, thanks.

Maybe it's better to do this in smaller steps than one big change.

following5:
> Regarding altitude: We could add that to the NPA cronjob

Looks like the NPA cronjob is out of service at OCPL sites. Lots of `caches.need_npa_recalc` = 1 in the develsite database.

Will make a proposal for an OKAPI-sync cronjob, which can do all those things like update merit badges, altitudes, etc. 

Btw, it just took half a minute on my develsite to update a cache altitude. If it's done by cronjob, the user interface is not blocked.

following5:
Added country and region editing and some supportive functions; now all fully reviewed and tested. This also allows easy implementation of searching by country, region and (description) language.

Description and region editing was more complicated than I expected ... but now works fine.

To do:
* https://github.com/opencaching/opencaching-pl/pull/1769
* editing of additional waypoints
* editing of images

following5:
This is ready for release, but the description editing may be improved. Currently the method allows to save description, short description and hint in just one language per call. This keeps the OKAPI interface and implementation rather simple, but complicates error handling for clients.

Alternatively, JSON dictionaries with multiple language for each of the three fields could be accepted. Or just one two-level dictionary, with language first, and the field names and content on second level. This would make the OKAPI interface and implementation way more complicated, but the usage cleaner.

How to proceed?

kojoty:
I understand cache edit by OKAPI is ready to go? 
@following5 do you need anything more to release it?

following5:
A decision needs to be made on the interface (and implementation) for description editing, see above. As Wrygiel is no longer engaged in Okapi and I am busy with some OCPL tasks, this will have to wait.

following5:
Also, this change exposes country and language name translations from OCPL database, which are partially broken (see https://github.com/opencaching/opencaching-pl/issues/1387#issuecomment-449685831). So it would be best to postpone it until either https://github.com/opencaching/opencaching-pl/issues/1387 or https://github.com/opencaching/okapi/issues/576 is done.

following5:
>  JSON dictionaries with multiple language for each of the three fields could be accepted.

I think we should go for this option. It gives the cleanest interface:
* Input to caches/edit has identical format as output of caches/geocache(s)
* error handling is not compromised
* no nested dictionaries

following5:
This change is blocked by several open issues:
* services/caches/capabilities must be rewritten
* OCPL capability settings must be recognized
* translations of OCPL country and language names should be improved
* description editing should be improved
* new OCPL cronjobs must be enabled at OC.RO (for altitude changes)

Also, attribute editing may be improved by adding an overwrite-by-new-set option, so developers don't need to calculate diffs for certain use cases.

following5:
`countries` table is no longer used by OCPL code: https://github.com/kojoty/opencaching-pl/commit/44f3ef39ec5023f43f3915df9463bbbfc828ad82

Issue #271, opened 2015-07-19
OC-specific attributes in GPX files


GoogleCodeExporter:
```
Excluding OC-specific attributes from GPX files is a severe restriction 
compared to the GPX format generated by OC.de, which includes all attributes 
(localID + 100 for only-oc attributes, and local English names). Generally it 
is a restriction of OC functionality and makes OC GPX export inferior to GC GPX 
export.

OKAPI returns local cache, user and log IDs in GPX files, so exposing internal 
IDs here obiously is no concern. Also, any application importing GPX files will 
ignore unknown attributes, as Groundspeak can add new attributes any time. 
There have never been any incompatibilities reported in all the years OC.de 
included own attribute IDs in GPX.

The only reason I can imagine for not including OC-specific attributes is that 
if we use makeshift attribute IDs (let's say OC-ID + 1000), Groundspeak may 
render them incompatible by intentionally introducing new attributes with the 
same value (current GC IDs range from 1 to 66). I think that this case is so 
VERY unlikely that we can safely implement this, at least as an alternative 
option to gs:attrs.

So I would like to add this feature.
```

Original issue reported on code.google.com by `following09` on 25 Jun 2013 at 6:54

PR #591, opened 2019-06-20
Refactoring attributes

andrixnet:
In reference to https://github.com/opencaching/opencaching-pl/issues/1251

attribute-definitions.xml has be substantially enhanced and synchronized with proper attribute IDs according to the changes to be implemented by https://github.com/opencaching/opencaching-pl/issues/1251 and missing attributes added. 

This work is only on attribute IDs, strings not (yet) changed (including translations). Only english strings to added attributes. 

Please review these changes to ensure correctness overall. 

Changes according to attributes documented here: https://wiki.opencaching.eu/index.php?title=Cache_attributes

Known possible side effects: 
- some handling of attributes may be affected (code mistakes to be corrected later)
- OC specific attributes handled by third party apps may not be listed correctly until third party apps update according to https://wiki.opencaching.eu/index.php?title=Cache_attributes
(AFAIK OC attributes have only partial support in at least c:geo)

Please schedule merging somtime between June 24 and 25th. 
It would be best to coordinate with corresponding merge at opencaching-pl.


okapi/services/attrs/attribute-definitions.xml
```diff
@@ -6,6 +6,12 @@ include any attribute used by at least one of the Opencaching installations.
 If you'd like to get the write access to our repository (to submit your
 translations), contact us: Submit a new issue in our issue tracker.
 
+DOCUMENTATION
+=============
+
+A comprehensive documentation of attribute assignments can be found at:
+https://wiki.opencaching.eu/index.php?title=Cache_attributes
+
```
following5: I will review that documentation.

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -3976,4 +4024,137 @@ for the original implementations.
         </lang>
     </attr>
 
+    <attr acode="A78" categories="de-cache-types">
+        <ocgs id="216" inc="true" />
```
following5: So far I have packed OCGS IDs into the numberspace 106-156, and would have used 159+ if none of those are left. (28, 129, 136, 138, 139, 140, 144, 146, 148, 151, 152 still had been free.)

I would prefer to keep the OCGS numberspace as compact as possible, because we are cannibalizing Groundspeak numberspace here. GS some time may need more then 105 different attributes, which would make new GS attrib IDs collide with OCGS IDs. Then we will need to negotiate with GS so that they skip OCGS numberspace. Negotiations may be easier if we did our best to keep OC footprint in their numberspace small.

(Of course, it would have been much better if the inventor of that OCGS number scheme would have started at 1000 rather than 100, avoiding collisions if Groundspeak adds some dozend new attribs. But he did not... and ended up with OCGS IDs in the range 106 - 158. So I stayed with that range for backward compatibility with OC.DE site.

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -206,10 +225,11 @@ for the original implementations.
     </attr>
 
     <attr acode="A2" categories="de-cache-types">
-        <ocgs id="117" inc="true" />
-        <opencaching schema="OCPL" id="54" />
-        <opencaching schema="OCNL" id="54" />
-        <opencaching schema="OCRO" id="54" />
+        <ocgs id="208" inc="true" />
```
following5: OCGS IDs must not be changed. This would break backward compatibility with applications that already rely on the existing IDs.

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -129,14 +141,20 @@ Attributes according to Geocaching.com:
 https://www.geocaching.com/about/icons.aspx
 gcTour GreaseMonkey script:
 https://gist.github.com/DieBatzen/5814dc7368c1034470c8/
+and based on:
+https://forums.geocaching.com/GC/index.php?/topic/282652-groundspeak-gpx/&tab=comments#comment-4855718
 
+**  This attribute is not user editable, instead used as maintenante flag.
```
bekuno: s/maintenante/maintenance/

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -50,9 +56,15 @@ Legend and field usage:
...
             (use <groundspeak ... name="NAME" .../>)
 
-Note: in Opencaching implementations all attributes are either ON or OFF.
-Inverse meaning is handled by a different attribute.
+Notes: 
+- In opencaching-pl implementation all attributes are either ON or OFF.
+  Inverse meaning is handled by a different attribute.
+- In OCDE implementation attributes have same ID and use:
+  YES <=> inc=1  
+  ON  <=> inc=0
```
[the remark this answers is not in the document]
andrixnet: Actually this is based directly on column `gc_inc` from 
https://github.com/OpencachingDeutschland/oc-server3/blob/development/sql/static-data/cache_attrib.sql

I presume OCDE code uses it only for it's own GPX exports.

okapi/services/attrs/attribute-definitions.xml
andrixnet: Will make the correction. Thank you.

okapi/services/attrs/attribute-definitions.xml
andrixnet: On the other hand, we could go directly to 1000 based OC attribs from here on (and also keeping the 9000 convention for negated attributes). This would make it very unlikely to overlap with GS IDs in the future.

Another important concept in bringing order to much existing almost chaos was to unify all [most] numberings. OKAPI has it's own aggregation and is also a GPX generator source (AFAIK). So ACODES are the second numbering. 

My proposal (next stage) was towards OCDE to switch to the same numberings. (AFAIK easier, since most changes happen in database and little in the code)
And to all third party apps to take into account the new and unified numbering scheme, for the first time (AFAIK) well documented. 

I am not sure [at this time] which/how many apps make actual use of OCGS elements in GPX to assess the potential impact.
I do know some apps (c:geo being one of them) make actual use of OKAPI ACODES.

A transitional period would be inevitable, with apps updating to the new numbering, leaving only old, unmaintained apps or outdated installations still using the previous numbering schemes.

okapi/services/attrs/attribute-definitions.xml
andrixnet: What would the actual real world impact (in the view of what is said in the comment just above)?

okapi/services/attrs/attribute-definitions.xml
andrixnet: Regarding potential negotiations: I seriously doubt GS would even take the time to listen...

okapi/services/attrs/attribute-definitions.xml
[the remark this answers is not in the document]
following5: Yes, that's for GPX exports.

okapi/services/attrs/attribute-definitions.xml
following5: Any application which processes OC specific attributes from GPX files will no longer recognize them. E.g. GASK will no longer show the OConly attribute, if 106 would be remapped.

following5: Also, OKAPI GPX would no longer be compatible with OCDE GPX.

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -4070,7 +4070,6 @@ store these attributes in GPX <groundspeak:attribute> elements.
 
     <attr acode="A82" categories="de-cache-types">
         <ocgs id="107" inc="true" />
-        <opencaching schema="OCDE" id="7" />
```
following5: I think A82 can be completely removed, as it is never used.

andrixnet: Oh. I see. Will remove.

okapi/services/attrs/attribute-definitions.xml
andrixnet: In light of this discussion I am strongly leaning towards making the ID redefinitions 1000 based.

okapi/services/attrs/attribute-definitions.xml
andrixnet: As I said before, I strongly recommend we urge third party app developers to adopt the new numbering scheme. 
AFAIK GSAK is a special case due to Clyde's illness.
Also AFAIK most popular geocaching apps that support OC have only partial support. 
Perhaps in part due to lack of documentation (until now?)

andrixnet: I am willing to submit a DB and code update to OCDE along these lines. 
Would OCDE be willing to merge such a pull request? 
Then GPX compatibility won't become a problem.

As you know I am trying to bring some order to a lot of chaos and whatever scenario one considers ofcourse there will be pros and cons. IMO the best solution is to have a well defined set of data, given the circumstances with no overlaps and by design ensuring minimal chances of future overlaps and have it uniformly deployed and well documented. Sometimes it is best to move on rather then clinging to an old and chaotic scheme that is a nightmare to maintain. 
Just as an example: the Linux kernel developers dropped i386 and i486 CPU support at some point.

andrixnet:
In the light of recent conversations, please don't commit yet. 
Reconsidering parts of the numbering scheme.

okapi/services/attrs/attribute-definitions.xml
following5: Backward compatibility has always been a core principle of OKAPI. We would never break backward compatibility unless we are absolutely forced to do so.

OCDE has been exporting those IDs for almost 10 years now. There is no need to change them and break apps. They don't need any mainenance. They don't need to be "beautiful" in any way. They have been defined once, and noone ever will have to care again about them. 

New attributes are added very, very rarely. For the case this is done, I already documented which OCGS IDs are free and can be used for new attribs. This is all documented [here](https://github.com/opencaching/opencaching-pl/blob/4415114673e8dc713d844236e970cf3f134defba/lib/format.gpx.inc.php#L342). OCGS IDs for the special UK and US attributes are also defined there. Please use those, instead of trashing al the work the I already did.

andrixnet: Understood. I will resequence the numbers with OCGS IDs as instructed.

However, with all due respect, as internal attrib IDs (independent from OCGS numbering) at each ocpl-based node will switch to a unified numbering and further attrib changes at UK, US, NL and RO (as requested by each), _some code will simply clean itself up_ because there won't be any further need to handle exceptions and exceptions to exceptions. It nothing about "trashing" anything and no ill intent.

andrixnet: Will return early next week with modified proposal.

following5:
It may be best if someone else from OCPL team takes over OKAPI maintenance. I have little time these days for OC, and therefore maybe I am not as open-minded and supportive as I should be for ongoing innovations at OCPL.

okapi/services/attrs/attribute-definitions.xml
andrixnet: Removed in next commit.

okapi/services/attrs/attribute-definitions.xml
andrixnet: Fixed in next commit.

okapi/services/attrs/attribute-definitions.xml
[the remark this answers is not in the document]
andrixnet: Changed the text to better explain the situation.

okapi/services/attrs/attribute-definitions.xml
andrixnet: As said, OCDE has been around for 10+ years, however the OCGS concept was introduced to OKAPI in 2017, so it is quite newer. 

I've checked the OKAPI code and, as far as I can tell so far, OCGS elements do not generate new elements in GPX export, instead they fill out <groundspeak:attribute elements and that's it. ( https://github.com/opencaching/okapi/issues/271 )
Same goes for OCDE GPX export and OCPL GPX export. 
I've also read several previous issues and discussions at OKAPI and gpx-extension-v1. 

AFAIK GSAK implements more extensively only OC-only (as attrib 106), including it in the filter dialog. Other OC specific attributes are displayed for the cache (taken from OCDE) but exported in GPX as ID only. 

My proposal does not make any changes to OC specific attributes that exist at OCDE (either both OCDE&OCPL or OCDE only). Thus any GPX exported by OKAPI or by OCDE code will not be affected. 

There seems to be very little support for OC attributes by third party apps and it seems to be more concentrated towards OCDE.

OCPL is in the long (and slow) process of implementing data definitions in the code to have a common base across installations (aiming to have in database only actual data, not data definitions; list of possible attributes counts as data definition). This cannot be done properly while each OCPL site has the same attribute with a different ID. 
Thus overhauling OCPL attributes is a critical step in that direction and a badly needed cleanup in a lot of chaotical data. It also simplifies the GPX export code.

My proposal changes only OCPL specific attributes with OCGS numbers in sync with OCPL internal ID numbers, assigned, I admit, to a different scheme then described here https://github.com/opencaching/opencaching-pl/blob/4415114673e8dc713d844236e970cf3f134defba/lib/format.gpx.inc.php#L342 ; 
Since even that numbering is non-continous, jumpy and with holes, in order to do the best to avoid further confusion, this proposal uses, for OCPL only, a distinct numbering set, compact, continuous, non-overlapping.

With OCGS IDs set as such and with OCPL code updates, with OCDE unaffected, GPX exports between OKAPI and OCPL code will also remain in sync, with same IDs.

I am also working with c:geo developers to extend their previously sketchy OC attribute support to improve it, greatly helped now by the documentation at https://wiki.opencaching.eu/index.php?title=Cache_attributes .

andrixnet:
Considering all that is said [here](https://github.com/opencaching/okapi/pull/591#discussion_r299038256) and given the fact that OCDE GPX vs OKAPI GPX is unaffected by this, please reconsider and allow OCPL to move forward. 
I do not know (at this time) if/when someone from OCPL might take over maintaining OKAPI. 
AFAIK wrygiel has left for some time now; my own knowledge of PHP is far too inadequate for that purpose anyway.

okapi/services/attrs/attribute-definitions.xml
following5: OCDE was founded in 2005. The OCGS concept was introduced at OCDE around 2010. It was added to OCPL in 2017, and to OKAPI in 2018.

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -57,11 +57,12 @@ Legend and field usage:
             (use <groundspeak ... name="NAME" .../>)
 
 Notes: 
-- In opencaching-pl implementation all attributes are either ON or OFF.
+- All OC implementations have attributes either ON or OFF.
   Inverse meaning is handled by a different attribute.
-- In OCDE implementation attributes have same ID and use:
+- In OCDE implements internally the additional `inc` flag for GS compatibility 
+  of it's GPX exports (for the representation of the inverse meaning of an attribute):
```
following5: Not only OCDE, but also OCPL and OKAPI implement the 'inc' flag for GPX exports.

okapi/services/attrs/attribute-definitions.xml
andrixnet: Hence it is quite new from OCPL perspective. Also attributes for OCPL sites in OKAPI were incomplete and inconsistent. 
Taking into consideration that this is the first time since 2014 (when I first got involved in OCPL) that attributes for OCPL sites in attribute-definitions.xml are 100% consistent (with this PR) and documented and that OCDE data is untouched, I respectfully request to allow OCPL to move forward.

okapi/services/attrs/attribute-definitions.xml
andrixnet: Improved notes on `inc` flag. Thank you.

okapi/services/attrs/attribute-definitions.xml
following5: If I shall do this, it needs more time. I try to 100% review and do some tests before merge. My OCPL devel VM is broken since a hard disk crash ~2 weeks ago, need to repair that first.

andrixnet: Thank you. It is very wise to do so. 
Please bear in mind that this PR is designed to be deployed together with this [PR](https://github.com/opencaching/opencaching-pl/pull/2069) at opencaching-pl which includes per-site database content updates.

andrixnet:
Getting ready to merge at opencaching-pl: https://github.com/opencaching/opencaching-pl/pull/2069
Please schedule a merge at OKAPI soon after. 

Thank you.

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -50,9 +56,23 @@ Legend and field usage:
...
+  * one via OC implementation's own code
+- OCDE implements internally the `inc` flag (in the database) for GS compatibility 
+  of it's own GPX exports (for the representation of the inverse meaning of an attribute):
+  YES <=> inc=1  
+  NO  <=> inc=0
+- OCPL generates the `inc` innternally in the code for GS compatibility of 
+  it's own GPX exports.
 
+Groundspeak attributes:
+---------------------------------------------------------------------------
```
bekuno: lineNumber: 75; columnNumber: 3; string "--" is not allowed in comments

okapi/services/attrs/attribute-definitions.xml
```diff
@@ -3976,4 +4032,126 @@ for the original implementations.
...
+        <lang id="en">
+            <name>NOT available during winter</name>
+            <desc>
+                NOT available during winter.
+            </desc>
+        </lang>
+    </attr>
+
+    <attr acode="A85" categories="de-cache-types">
+        <groundspeak id="1" inc="false" name="Dogs" />
```
bekuno: Should GS ID 1 (dogs) really map to A85 (no dogs)?
This is also not so described in the wiki entry.

bekuno: Ok, this is a negative ID. 
But then your wiki page had be to changed for ID 9013 (remove gc_id 13).

andrixnet: A85 corresponds to GC id=1 with inc=0 (no dogs allowed, negative meaning)

andrixnet: Wiki page is correct regarding A40, the internal implementation at OCDE has separate columns regarding `gc_id` and `inc`.

okapi/services/attrs/attribute-definitions.xml
andrixnet: removed.

okapi/services/attrs/attribute-definitions.xml
bekuno: Not to me.

Please compare the state of ID 14/9014 with ID 13/9013 in the Wiki table and attribute-definitions.txt. 
In the file attribute-definitions.txt the characters 41 and 42 have both entries <groundspeak id="14".....
and acode 39 and 40 have <groundspeak id="13" .....

So I think that the documentation in the wiki should only be implemented with one kind and not with two different ones.

andrixnet:
```
    <attr acode="A40" categories="de-time-and-seasons">
        <groundspeak id="13" inc="false" name="Available at all times" />
        <opencaching schema="OCDE" id="39" />
        <opencaching schema="OCNL" id="9013" />
        <opencaching schema="OCRO" id="9013" />
        <opencaching schema="OCUK" id="9013" />
        <opencaching schema="OCUS" id="9013" />
```
Wiki:
`9013 | NOT available 24/7 | Available only during open hours | 39 | 13 | time | Only available at specified times | 0 | A40`

-----------------------------
Then: 
```
    <attr acode="A41" categories="de-time-and-seasons">
        <groundspeak id="14" inc="false" name="Recommended at night" />
        <opencaching schema="OCDE" id="40" />
        <opencaching schema="OCUK" id="9014" />
```
Wiki:
`9014 | Day | NOT recommended at night | 40 | 14 | day | By day only | 0 | A41 `

-----------------------------
Then: 
```
    <attr acode="A42" categories="de-time-and-seasons">
        <groundspeak id="14" inc="true" name="Recommended at night" />
        <opencaching schema="OCPL" id="14" />
        <opencaching schema="OCNL" id="14" />
        <opencaching schema="OCRO" id="14" />
```

Wiki:
`14 | Night | Recommended at night |   |   |   |   |   | A42`

-----------------------------

Which is correct as intended. 
- **A39** is Groundspeak attr 13; **inc=true** (Available 24/7)
- **A40** is Groundspeak attr 13; **inc=false** (NOT available 24/7)
 with OCDCE data being: internal ID=39, gc_id=13, inc=0 (false)
and OCPL data being: internal ID=9013.

OC attributes are not 3-state like GC attributes. Negating an attribute is represented by a standalone attribute whose meaning is the negation of another attribute. 
OKAPI does it's best to include compatibility information such that each OC attribute (where possible) may be represented by it's Groundspeak equivalent.
With this in mind, it is perfectly normal for A39 and A40 to reference the same Groundspeak attribute ID (13 in this case).
Same logic applies to A41 and A42.

Furthermore, since OCDE and OCPL implementations differ, OKAPI also does it's best to present a unified API and a unified view. 
For all intents and purposes, third party apps using OKAPI should use the ACODES to represent attributes. Using the `<groundspeak>` tag being for compatibility with existing app code and data to handle Groundspeak cache information.  All the rest are **internals** should the app need or desire to present such detailed differences. 

The logic is explained in the section [Attribute_numbering_rules](https://wiki.opencaching.eu/index.php?title=Cache_attributes#Attribute_numbering_rules) from the Wiki.

kojoty:
@following5, hi are you there? I need your opinion about this issue. Thanks!

following5:
Sorry for my long silence.

I have reviewed this change again and can confirm, that
* it runs stable (will not crash OKAPI)
* it does not affect OCDE
* the changed GC attribute names are ok,
* it breaks backward compatibility of Groundspeak attribute IDs for OCPL-only attributes

The latter one means that if developers are using the ` gc_ocde:attrs` option of [services/caches/formatters/gpx](https://opencaching.pl/okapi/services/caches/formatters/gpx.html) and rely on the IDs which are currently returned, the OCPL-specific attributes will become "invisible" for these apps until developers implement an update and users install that update.

So far OKAPI has promised developers to always stay backward compatible. This would be the first case when we deliberately break backward compatibility. But as this feature has been enabled for OCPL sites just one year ago and is rather "exotic", it *may* not be in use yet in any applications. Anyway, though I am not comfortable with this change (and think it is unnecessary), I will leave the decision to implement this up to OCPL developers.

I did not review the changed and new OCPL attribute definitions, due to a lack of time. If you think they are ok, please feel free to merge this PR.

Note that at least the following changes need to be done at OCPL sites **as synchronously as possible**, so that OKAPI attributes continue to work properly:

* update the mapping table in lib/format.gpx.inc.pcp of OCPL code, lines 267 to 393
* update the mapping table in lib/search.gpx.inc.php of OCPL code, lines 185 to 224. Note that ther are *two* GPX implementations in OCPL, this is the older one, but still in use.
* update the IDs in the `caches_attributes` tables of all OCPL sites via some database update.

Also, the `attribute_` IDs in `lib/languages/*.php` need to be updated for all languages **and uploaded to Crowdin** (otherwise all the attribute translations at Crowdin would get lost). These translations are not in use yet, but are needed to eliminate the `cache_attrib` table.

The renumbering of IDs is much work, it is complicated to deploy, may introduce new bugs and will not bring benefits to users or developers. It is a purely cosmetical change. Therefore I do not think this is a good idea. But again, if OCPL developers want to do it, it's your decision.

following5:
> These translations are not in use yet

or at least were not in use when I defined them some time ago. Don't know what happened since then in OCPL code.

wrygiel:
I didn't read the discussion here, nor reviewed the changes, but I share with @following5's doubts. Be sure to backup all databases before this change, because there's a significant chance that this backfires.

Issue #594, opened 2019-07-04
OCPL: Cache types


andrixnet:
With new additions and modifications at opencaching-pl regarding cache types, OKAPI needs to be updated as well. 

See https://wiki.opencaching.eu/index.php?title=Cache_types for documentation.
See opencaching-pl [New cache types](https://github.com/opencaching/opencaching-pl/issues/860)
See opencaching-pl [CACHE_TYPE_ID changes](https://github.com/opencaching/opencaching-pl/issues/2024)
See opencaching-pl [GeoPath FINAL cache type](https://github.com/opencaching/opencaching-pl/issues/2076)

andrixnet:
Please review proposed changes here: https://github.com/andrixnet/okapi/commit/2d868b99805b5d1f5d61fa8084554cd219a01e60
- added/changed OCPL cache id (internal)
- added new/missing graphical resources

AFAIK there should not be any backwards compatibility issues since cache types are exposed by name, not by internal IDs. 

Note: /okapi/services/caches/geocache.html documentation not (yet) updated.

Thank you.

following5:
> AFAIK there should not be any backwards compatibility issues since cache types are exposed by name, not by internal IDs.

For OKAPI it does not matter, but the OCPL and OCDE [XML Interface](https://github.com/opencaching/opencaching-pl/tree/master/xml) exposes internal IDs. This interface may still be used e.g. by GeoKrety.

andrixnet:
Thank you for pointing that out. 
Geokrety has been notified [here](https://github.com/geokrety/geokrety-website/issues/427)

andrixnet:
Corrections made [here](https://github.com/andrixnet/okapi/commit/2382f7aa743794cfc24684421e42a3b0f99af8ce) and documentation added.

andrixnet:
GK is working on switching to OKAPI: https://github.com/geokrety/geokrety-scripts/issues/9 so the old XML interface may indeed become obsolete.

following5:
> GK is working on switching to OKAPI: geokrety/geokrety-scripts#9 so the old XML interface may indeed become obsolete.

It looks like XML interface has been removed from OCPL code.

PR #595, opened 2019-07-23
Cache type changes at OCPL

andrixnet:
Handles changes described here: https://github.com/opencaching/okapi/issues/594 and is paired with this PR at OCPL: https://github.com/opencaching/opencaching-pl/pull/2087


okapi/services/caches/geocache/docs.xml
```diff
@@ -164,13 +164,20 @@
...
+                        of goal or task,</li>
+                    <li><b><s>Podcast</s></b>, a geocache with attached MP3 file(s). 
+                        <i>This cache type no longer exists and it's reference is kept
+                        for compatibility reasons.</i></li>
                     <li>(Types may be added or removed. Your application MUST be
                         prepared for any new types and may handle them as "Other".)</li>
                 </ul>
-                <p>There are even more types, but OKAPI maps them to common types
+                <p>Opencaching cache types are documented 
+                <a href="https://wiki.opencaching.eu/index.php?title=Cache_types" target="_blank">here</a></p>
```
following5: Everthing OKAPI developers need to know should be documented *here*, so it is not necessary to reference any external documentation. To avoid redundancy and all its inherent problems, I suggest to remove this reference from OKAPI docs.

following5:
This PR changes the internal OCPL ID of Moving caches from 8 to 9. Please be aware of the consequences of such renumbering:

* OCPL code needs to be examined carefully for any references to cache type 8, changing it to 9. There are many places in OC code, including templates and translation tables, where cache type IDs are referenced.
* Synchronous update of OKAPI, OCPL code and contents of 'caches'  and 'cache_types' tables must be done on all OCPL sites.

I don't see any benefit by this renumbering and recommend not to do it - it just produces additional work and can introduce bugs.

andrixnet:
I need to review this again, if still pertinent or close.

PR #596, opened 2019-10-31
A parameter awarded_status added to cache search filter

rapotek:
Extends the cache search filter parameters by an `awarded_status` one, which includes the `cache_titled` table in the search query, existing in opencaching-pl only. The term "awarded" used here seems to be more correct than "titled" used in table name.

Available values (see docs): `awarded_only`, `notawarded_only`, `either`.
The code formatting is based on the existing one in the same method.

### DIFF
--- a/okapi/services/caches/search/all/docs.xml
+++ b/okapi/services/caches/search/all/docs.xml
@@ -262,19 +262,20 @@
         IDs. You should not use them. This parameter was implemented for
         internal use.</p>
     </opt>
-    <opt name='awarded_status' default='either'>
-        <p>Includes information about a "cache of the week/month etc." award
-        granted or not in the search criteria. Only some Opencaching sites
-        implement this feature. If you set the parameter to one of the following
-        values and this site does not implement it, it will not affect the
-        search result.</p>
+    <opt name='awarded'>
+        <p>Pipe separated list of award types. If given and not empty, an award
+        of selected type is added to filter constraints if it is applicable to
+        current Opencaching site. If a '-' character precedes the award type,
+        only caches <b>not granted</b> corresponding award will be included
+        in the resulting set. If a '+' character precedes the award type or
+        there is no preceding character, only caches <b>granted</b>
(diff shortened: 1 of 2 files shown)


following5:
This looks formaly ok, but there may be an issue with the parameter definition.

Let's assume another type of award is implemented in the future, something completely different than the cache of the week/month/year awards. Should the option `awarded_only` then select only caches which got the cache of the week/month/year award, or should it select any caches that got an award?

Depending on the answer, it may be better to either chose a more specific name for the new search option or to defini it more general.

okapi/services/caches/search/all/docs.xml
```diff
@@ -262,6 +262,21 @@
         IDs. You should not use them. This parameter was implemented for
         internal use.</p>
     </opt>
+    <opt name='awarded_status' default='either'>
+        <p>Includes information about a "cache of the week/month etc." award
```
wrygiel: "Includes" suggests that using this parameter includes something into the response. It's actually the opposite. I'd suggest some rewording, e.g. "Allows to filter caches based on awards it has been granted."

following5:
In the other include/exclude-list search options - `type`, `size2` and `owner_uuid` - there is no prefix for including, only the "-" for excluding. I think we should keep a consistent syntax here and not accept a + prefix. No "-" means include.

following5:
The SQL query will crash if a user supplies duplicate awards, e.g. `awarded=cotp|cotp` or `awarded=cotp|-cotp` .

following5:
I think it should return an empty set of caches in these two cases:
* `awarded=cotp|-cotp` on OCPL sites
* `awarded=cotp` on OCDE sites

following5:
>*  awarded=cotp|-cotp on OCPL sites
>*  awarded=cotp on OCDE sites

... and of course any equivalent request which adds more `awarded` options. Empty on OCPL if both include and exclude are present, empty on OCDE if include is present.

PR #598, opened 2020-06-08
Blacklist for okapi cron jobs

kojoty:
Now it is possible to temporary disable some okapi cron jobs.
This should be used only for debug purpose in production environment.
No OKAPI cron jobs should be disabled permanently.


kojoty:
at OCPL we saw issues around:
- fullDump
- tileTreeUpdate 
jobs. This options allow to temporary disable some job to help debug the problems when job makes the server not working.

PR #604, opened 2021-03-15
fix for PHP-7.4

kojoty:
It seems previous code generates the error:
"Trying to access array offset on value of type null" if there is no $token in DB.


kojoty:
the same was in #605 :)

Issue #619, opened 2022-01-10
imagegd2(): GD2 image support has been disabled


andrixnet:
After a recent GD library debacle where some errors were being generated due to a bug in libgd-2.3.0 I managed to upgrade to libgd-2.3.3 on OCRO/NL/US. 

After the upgrade I started getting the following in error.log:
```imagegd2(): GD2 image support has been disabled\n, referer: https://www.opencaching.us/MainMap/fullScreen?lat=32.55035&lon=-92.165183333333&openPopup```

According to https://github.com/libgd/libgd/issues/428 GD2 format has been designated for development and testing purposes only and disabled by default. 

Obviously my upgrade was built according to default settings. 

Since this has been slated for being phased out of production anyway. 

AFAIK it seems to be used in TileRenderer.php for caching purposes. Perhaps something different could be used?

deg-pl:
@BonEchoGeocacher @TermiteHunter let's continue discussion from https://github.com/opencaching/opencaching-pl/issues/2344 here, because this is an OKAPI problem.

With @kojoty we done research, and this issue we'll fix next week.

kojoty:
The fix seems to be ready. I'm going to deliver it early on Monday.

BonEchoGeocacher:
The cache map is working again on OCNA. Many thanks to [kojoty](https://github.com/kojoty) and [deg-pl](https://github.com/deg-pl) for fixing this issue - very much appreciated!

andrixnet:
Many thanks.

PR #620, opened 2022-02-06
 #619: imagegd2 is no used as it is removed from latest libgd

kojoty:
Instead of gd2 png is used as an cached imgage.
  Also pixel by pixel opacity procedure is refactored to use generic function.

### DIFF
--- a/okapi/services/caches/map/TileRenderer.php
+++ b/okapi/services/caches/map/TileRenderer.php
@@ -15,7 +15,7 @@ class TileRenderer
      * Changing this will affect all generated hashes. You should increment it
      * whenever you alter anything in the drawing algorithm.
      */
-    private static $VERSION = 61;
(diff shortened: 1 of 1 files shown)


kojoty:
@teiling88 @deg-pl 
Please review and verify if it works for for you. I'm not sure of the performance of this but I assume there is no choice. I would like to deliver it ASAP as OCUS/OCRO/OCNL are affected by this issue - map doesn't work there...

teiling88:
Seems to be ok for oc.de

teiling88 (approved):

deg-pl (approved):
Looks good.
If you will change all $gd2* variables to other names ($png* for example) - will be perfect ;)

Yes, I know. I have little Asperger syndrome.

deg-pl:
Works great! Thanks @kojoty!

PR #621, opened 2022-11-23
NTR - remove group_contact to prevent string cutting for private notes

teiling88:
We have a strange issue in the c:geo application. The group_concat is cutting private notes cause of mysql default string limitations in group_contact. We don't need this cause we have no duplicate entries in coordinates.

For more details please see https://github.com/cgeo/cgeo/issues/13651

### DIFF
--- a/okapi/services/caches/geocaches/WebService.php
+++ b/okapi/services/caches/geocaches/WebService.php
@@ -1117,7 +1117,7 @@ class WebService
                 # OCDE uses coordinates table (with type == 2) to store notes (this is somewhat weird).
 
                 $rs = Db::query("
-                    select cache_id, null as date, group_concat(description) as `desc`
+                    select cache_id, null as date, description as `desc`
                     from coordinates
                     where
                         type = 2  -- personal note


kojoty (approved):
OK, I don't understand why group_concat can cause the issue but this is ocde specific so let's merge it

bekuno:
> OK, I don't understand why group_concat can cause the issue but this is ocde specific so let's merge it

see 
- https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_group_concat_max_len
- https://sebhastian.com/mysql-group_concat-limit/

kojoty:
> > OK, I don't understand why group_concat can cause the issue but this is ocde specific so let's merge it
> 
> see
> 
> * https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_group_concat_max_len
> * https://sebhastian.com/mysql-group_concat-limit/

Not sure if it is worth the discussion - I assume you know what you need anyway :)

I understand that group_concat has limit so if there are duplicated records descriptions of all such records will be concatenated but result could be no longer than mysql-group_concat-limit - this is clear 

I didn't understand why it cut the notes - I assumed this is not in the same column so why description has anything to notes?

but please note I didn't study the whole story :)

murggel:
group_concat is used for OCPL as well? Can the problem occur there as well?
https://github.com/teiling88/okapi/blob/0d6a2983dc5580168003b9733574fcedb15912cc/okapi/services/caches/geocaches/WebService.php#L1107


> I didn't understand why it cut the notes - I assumed this is not in the same column so why description has anything to notes?

For me it seems that the content of the note is saved in the description field...

teiling88:
image
The description field with type 2 is the private note in the ocde branch. This issue could happen in ocpl as well.

rapotek:
> group_concat is used for OCPL as well. Can the problem occur there as well?
> 
> https://github.com/opencaching/okapi/blob/233361623d6a4bcc7790f40a8164b14df1f7fc66/okapi/services/caches/geocaches/WebService.php#L1107
> 
> > I didn't understand why it cut the notes - I assumed this is not in the same column so why description has anything to notes?
> 
> For me it seems that the content of the note is saved in the description field...

`group_concat` can be safely removed here, too; `cache_notes` table has a primary key of (`cache_id`, `user_id`), so for single user there cannot be more than one record with the same `cache_id`. The grouping here is on `cache_id` so the result would be the same as for direct select of `desc` without `group_concat`.

Issue #628, opened 2023-11-23
New service method: save_user_coords


hxdimpf:
This is about adding a new service endpoint. The  platforms (opencaching.xx) provide the capabilities to maintain a personal cache notes field for each cache. In  addition with this there is also a capability to store user provided coordinates for instance the coordinates that are the result of solving a puzzle.

The OKAPI provides a service to retrieve the personal caches notes and also the user_coords. It also provides a service to update the personal cache notes but **not**  the user coords.

I have implemented a new service to fill this gap. First I thought about extending the already existing service

`:: services/caches/save_personal_notes` 

however, to make sure that this doesn't create any adverse side effects for existing client applications I decided to implement a new service:
`:: services/caches/save_user_coords`.

This issue will be referenced in an upcoming PR

PR #629, opened 2023-11-23
issue #628 save_user_coords

hxdimpf:
This is about adding a new service endpoint. The platforms (opencaching.xx) provide the capabilities to maintain a personal cache notes field for each cache. In addition with this there is also a capability to store user provided coordinates for instance the coordinates that are the result of solving a puzzle.

The OKAPI provides a service to retrieve the personal caches notes and also the user_coords. It also provides a service to update the personal cache notes but not the user coords.

I have implemented a new service to fill this gap. First I thought about extending the already existing service

:: services/caches/save_personal_notes

however, to make sure that this doesn't create any adverse side effects for existing client applications I decided to implement a new service:
:: services/caches/save_user_coords.

### DIFF
--- a/okapi/services/caches/save_user_coords/WebService.php
+++ b/okapi/services/caches/save_user_coords/WebService.php
@@ -51,58 +51,94 @@ class WebService
         );
         $cache_id = $geocache['internal_id'];
 
-        self::update_notes($cache_id, $request->token->user_id, $latitude, $longitude);
-
-	$ret_value = 'ok';
+        self::update_coordinates($cache_id, $request->token->user_id, $latitude, $longitude);
 
         $result = array(
-            'status' => $ret_value
+            'success' => true
         );
         return Okapi::formatted_response($request, $result);
     }
 
-    private static function update_notes($cache_id, $user_id, $latitude, $longitude)
+    private static function update_coordinates($cache_id, $user_id, $latitude, $longitude)
     {
-        /* See:
-         *
-         * - https://github.com/OpencachingDeutschland/oc-server3/tree/development/htdocs/src/Oc/Libse/CacheNote
-         * - https://www.opencaching.de/okapi/devel/dbstruct
-         */
+        if (Settings::get('OC_BRANCH') == 'oc.de')
+        {
 
-        $rs = Db::query("
-            select max(id) as id
-            from coordinates
-            where
-                type = 2  -- personal note
-                and cache_id = '".Db::escape_string($cache_id)."'
-                and user_id = '".Db::escape_string($user_id)."'
-        ");
-        $id = null;
-        if($row = Db::fetch_assoc($rs)) {
-            $id = $row['id'];
-        }
-        if ($id == null) {
-            Db::query("
-                insert into coordinates (
-                    type, latitude, longitude, cache_id, user_id
-                ) values (
-                    2,
-                    '".Db::escape_string($latitude)."',
-                    '".Db::escape_string($longitude)."',
-                    '".Db::escape_string($cache_id)."',
-                    '".Db::escape_string($user_id)."'
-                )
-            ");
-        } else {
-            Db::query("
-                update coordinates
-                set latitude  = '".Db::escape_string($latitude)."',
-                    longitude = '".Db::escape_string($longitude)."',
+            /* See:
+             *
+             * - https://github.com/OpencachingDeutschland/oc-server3/tree/development/htdocs/src/Oc/Libse/CacheNote
+             * - https://www.opencaching.de/okapi/devel/dbstruct
+             */
+
+            $rs = Db::query("
+                select max(id) as id
+                from coordinates
                 where
-                    id = '".Db::escape_string($id)."'
-                    and type = 2
+                    type = 2  -- personal note
+                    and cache_id = '".Db::escape_string($cache_id)."'
+                    and user_id = '".Db::escape_string($user_id)."'
             ");
+            $id = null;
+            if($row = Db::fetch_assoc($rs)) {
+                $id = $row['id'];
+            }
+            if ($id == null) {
+                Db::query("
+                    insert into coordinates (
+                        type, latitude, longitude, cache_id, user_id, description
+                    ) values (
+                        2,
+                        '".Db::escape_string($latitude)."',
+                        '".Db::escape_string($longitude)."',
+                        '".Db::escape_string($cache_id)."',
+                        '".Db::escape_string($user_id)."',
+                        '".Db::escape_string("")."'
+                    )
+                ");
+            } else {
+                Db::query("
+                    update coordinates
+                    set latitude  = '".Db::escape_string($latitude)."',
+                        longitude = '".Db::escape_string($longitude)."'
+                    where
+                        id = '".Db::escape_string($id)."'
+                        and type = 2
+                ");
+            }
+        }
+        else # oc.pl branch
+        {
+            $rs = Db::query("
+                select max(id) as id
+                from cache_mod_cords
+                where
+                    cache_id    = '".Db::escape_string($cache_id)."'
+                    and user_id = '".Db::escape_string($user_id)."'
+            ");
+            $id = null;
+            if($row = Db::fetch_assoc($rs)) {
+                $id = $row['id'];
+            }
+            if ($id == null) {
+                Db::query("
+                    insert into cache_mod_cords (
+                        cache_id, user_id, latitude, longitude
+                    ) values (
+                        '".Db::escape_string($cache_id)."',
+                        '".Db::escape_string($user_id)."',
+                        '".Db::escape_string($latitude)."',
+                        '".Db::escape_string($longitude)."'
+                    )
+                ");
+            } else {
+                Db::query("
+                    update cache_mod_cords
+                    set latitude  = '".Db::escape_string($latitude)."',
+                        longitude = '".Db::escape_string($longitude)."'
+                    where
+                        id = '".Db::escape_string($id)."'
+                ");
+            }
         }
     }
-
 }
--- a/okapi/services/caches/save_user_coords/WebService.php
+++ b/okapi/services/caches/save_user_coords/WebService.php
@@ -1,0 +1,108 @@
+<?php
+
+namespace okapi\services\caches\save_user_coords;
+
+use okapi\core\Db;
+use okapi\core\Exception\ParamMissing;
+use okapi\core\Exception\InvalidParam;
+use okapi\core\Okapi;
+use okapi\core\OkapiServiceRunner;
+use okapi\core\Request\OkapiInternalRequest;
+use okapi\core\Request\OkapiRequest;
+use okapi\Settings;
+
+class WebService
+{
+    public static function options()
+    {
+        return array(
+            'min_auth_level' => 3
+        );
+    }
+
+    public static function call(OkapiRequest $request)
+    {
+
+        $user_coords = $request->get_parameter('user_coords');
+        if ($user_coords == null)
+            throw new ParamMissing('user_coords');
+        $parts = explode('|', $user_coords);
+        if (count($parts) != 2)
+            throw new InvalidParam('user_coords', "Expecting 2 pipe-separated parts, got ".count($parts).".");
+        foreach ($parts as &$part_ref)
+        {
+            if (!preg_match("/^-?[0-9]+(\.?[0-9]*)$/", $part_ref))
+                throw new InvalidParam('user_coords', "'$part_ref' is not a valid float number.");
+            $part_ref = floatval($part_ref);
+        }
+        list($latitude, $longitude) = $parts;
+
+        # Verify cache_code
+
+        $cache_code = $request->get_parameter('cache_code');
+        if ($cache_code == null)
+            throw new ParamMissing('cache_code');
+        $geocache = OkapiServiceRunner::call(
+            'services/caches/geocache',
+            new OkapiInternalRequest($request->consumer, $request->token, array(
+                'cache_code' => $cache_code,
+                'fields' => 'internal_id'
+            ))
+        );
+        $cache_id = $geocache['internal_id'];
+
+        self::update_notes($cache_id, $request->token->user_id, $latitude, $longitude);
+
+	$ret_value = 'ok';
+
+        $result = array(
+            'status' => $ret_value
+        );
+        return Okapi::formatted_response($request, $result);
+    }
+
+    private static function update_notes($cache_id, $user_id, $latitude, $longitude)
+    {
+        /* See:
+         *
+         * - https://github.com/OpencachingDeutschland/oc-server3/tree/development/htdocs/src/Oc/Libse/CacheNote
+         * - https://www.opencaching.de/okapi/devel/dbstruct
+         */
+
+        $rs = Db::query("
+            select max(id) as id
+            from coordinates
+            where
+                type = 2  -- personal note
+                and cache_id = '".Db::escape_string($cache_id)."'
+                and user_id = '".Db::escape_string($user_id)."'
+        ");
+        $id = null;
+        if($row = Db::fetch_assoc($rs)) {
+            $id = $row['id'];
+        }
+        if ($id == null) {
+            Db::query("
+                insert into coordinates (
+                    type, latitude, longitude, cache_id, user_id
+                ) values (
+                    2,
+                    '".Db::escape_string($latitude)."',
+                    '".Db::escape_string($longitude)."',
+                    '".Db::escape_string($cache_id)."',
+                    '".Db::escape_string($user_id)."'
+                )
+            ");
+        } else {
+            Db::query("
+                update coordinates
+                set latitude  = '".Db::escape_string($latitude)."',
+                    longitude = '".Db::escape_string($longitude)."',
+                where
+                    id = '".Db::escape_string($id)."'
+                    and type = 2
+            ");
+        }
+    }
+
+}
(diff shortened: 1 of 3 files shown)


okapi/services/caches/save_user_coords/docs.xml
```diff
@@ -0,0 +1,30 @@
+<xml>
+    <brief>Update personal coordinates of a geocache</brief>
+    <issue-id>629</issue-id>
+    <desc>
+        <p>This method allows your users to to change the content of the user's
+        personal cache coordinates from one value to the other value.</p>
```
wrygiel, line 6: How about this:

This method allows your users to update the coordinates of their personal geocache coordinates.

okapi/services/caches/save_user_coords/docs.xml
```diff
@@ -0,0 +1,30 @@
...
+
+        <p>Current personal coordinates for the geocache can be retrieved
+        using the <b>alt_wpts</b> field in the
+        <a href="%OKAPI:methodargref:services/caches/geocache#fields%">services/caches/geocache</a>
+        method.</p>
```
wrygiel, line 11: The docs for this field say that "this type of waypoint is available only in some installations". Please explain how this method will behave in case of installations which do not support personal coordinates.

wrygiel (changes requested):

okapi/services/caches/save_user_coords/docs.xml
hxdimpf: good suggestion, will take it.

okapi/services/caches/save_user_coords/docs.xml
hxdimpf: I have checked `opencaching.pl`, it has the same database table so the code will work. Whether `opencaching.pl` renders this detail at the UI, I don't know. Am I supposed to check all other installations too?

I don't know how I could verify the code changes on a different installation.

okapi/services/caches/save_user_coords/docs.xml
```diff
@@ -0,0 +1,30 @@
+<xml>
+    <brief>Update personal coordinates of a geocache</brief>
```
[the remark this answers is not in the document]
hxdimpf: I agree. Would it make sense to make `cache_code` actually a "|" separated list of `cache codes` and `user_coords` a matching list (by position) of coordinates, such that only one request needs to be made for potentially a large set of caches? Such scenarios are quite common on some alternate geocaching platforms but I haven't seen a use case on opencaching.de as there are very few powertrails which would benefit from such a functionality.

Alternatively I have thought a about making this an additional, optional parameter to `save_personal_note`, however, in the interest of guaranteeing to not break existing client applications I decided against it.

With respect to your question: I don't think many more such single property changing methods are required. So I'd consider this a one of a kind. On my roadamp there is at this point a method let me call it "upload_fieldnotes", an issue for that exists and it has been discussed for years.

Other than that, if I can afford the time, I would like to add a full circle CRUD functionality for "lists". I can provide more details via a separate issue.

okapi/services/caches/save_user_coords/docs.xml
```diff
@@ -0,0 +1,30 @@
...
+	<p>Use positive numbers for latitudes in the northern hemisphere and longitudes
+	in the eastern hemisphere (and negative for southern and western hemispheres
+	accordingly). These are full degrees with a dot as a decimal point (ex. "48.7|15.89").</p>
+    </req>
+    <common-format-params/>
+    <returns>
+        <p>A dictionary of the following structure:</p>
+        <ul>
+            <li>status - ok</li>
+            <li>user_coords - the updated values</li>
```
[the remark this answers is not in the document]
hxdimpf: Yes, riight. What type of response would you suggest in such a case? Just "nothing" or would be a "status : ok" be sufficient?

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,105 @@
...
+        # Verify cache_code
+
+        $cache_code = $request->get_parameter('cache_code');
+        if ($cache_code == null)
+            throw new ParamMissing('cache_code');
+        $geocache = OkapiServiceRunner::call(
+            'services/caches/geocache',
+            new OkapiInternalRequest($request->consumer, $request->token, array(
+                'cache_code' => $cache_code,
+                'fields' => 'my_notes|internal_id'
```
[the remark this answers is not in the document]
hxdimpf: That is a leftover from copy/paste I will remove it right away.

okapi/services/caches/save_user_coords/docs.xml
wrygiel: You need to check only OCDE and OCPL only, all the other ones use the code from one of these. We can add reviewers from OCPL later on, or you can ask for them now (email rt@).

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,105 @@
...
+        }
+        if ($id == null) {
+            Db::query("
+                insert into coordinates (
+                    type, latitude, longitude, cache_id, user_id,
+                    description
+                ) values (
+                    2, $latitude, $longitude
+                )
+            ");
```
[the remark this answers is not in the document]
hxdimpf: arent the remaining fields filled with defaults? I will check and set meainingful values in the other 3. I must say I am not an SQL guy, I use MongoDB and it is no problem there to do things I did above.

The values $latitude, $logitude are not strings, those are floats, hwo would "Db::escape_string" help here?

okapi/services/caches/save_user_coords/docs.xml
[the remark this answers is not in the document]
wrygiel: We already have methods which [retrieve fields from multiple entities](https://github.com/opencaching/okapi/assets/2168535/311812d5-e081-4304-a550-e59700191be0), and corresponding methods which retrieve fields from a single entity (e.g. geocache, entry, etc.). Since that's the style already, we should follow the same pattern with saving methods.

So, we could have:

- `update_geocache` which would take `cache_code` and fields such as `personal_note`, `user_coords`, `marked_as_watched`, `marked_as_ignored`, `passwd`, etc. (This method would replace `edit`, `mark`, `save_personal_notes` and `save_user_coords`.)
- `update_geocaches` which would take a JSON list of arguments for `update_geocache`, similar to what `search_and_retrieve` does.

> With respect to your question: I don't think many more such single property changing methods are required. So I'd consider this a one of a kind.

It's fine to leave it as is then. We can always clean it up later (in a backward-compatible way).

okapi/services/caches/save_user_coords/docs.xml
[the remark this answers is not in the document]
wrygiel: This case looks very similar to the [mark method](https://www.opencaching.de/okapi/services/caches/mark.html), so I'd return the same format here.

okapi/services/caches/save_user_coords/WebService.php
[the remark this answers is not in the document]
hxdimpf: Pls re-review. Is this better now?

hxdimpf (commented):
Have implemented all suggested changes

okapi/services/caches/save_user_coords/docs.xml
hxdimpf: Ok, I just reviewd the database layout for opencaching.pl and as it seems, if this is the right place: `https://opencaching.pl/okapi/devel/dbstruct` the OCPL does not have a `user_coord` field in the respective table, which has even a different name and different fields. Now the question what to do here? I could make the implementation conditional to OCDE and do nothing for OCPL but the whole method should then not even show up for the OCPL installation in which case the conditional implementation wouldn't be needed alltogether.

Could the service be conitionally added to the service runner such that it doesn't even show up on the list for the OCPL installation and also not on their documentation?

okapi/services/caches/save_user_coords/WebService.php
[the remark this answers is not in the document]
wrygiel: Yes, just surround latitude and longitude with escape_string too (as a general rule we do this for all SQL-injected values, regardless of their type).

okapi/services/caches/save_user_coords/docs.xml
wrygiel: You could add a method with `ocde-specific` infotag (see the docs for `services/apiref/method`). But let's first confirm with OCPL guys if indeed OCPL doesn't keep this data. It's possible that OCPL still has it, just in some other table/field. If that's the case, then we should support it.

@opencaching/opencaching-pl-lead-programmers Can you comment?

okapi/services/caches/save_user_coords/WebService.php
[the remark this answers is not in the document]
wrygiel: MySQL will accept quoted and escaped `'123'` as well as `123`.

hxdimpf (commented):
did all the string escapes as suggested

okapi/services/caches/save_user_coords/docs.xml
hxdimpf: I looked at the dbstruct for OCPL and the following seems to be the right place:

```
DROP TABLE IF EXISTS `cache_mod_cords`;
/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!40101 SET character_set_client = utf8 */;
CREATE TABLE `cache_mod_cords` (
  `cache_id` int(11) NOT NULL,
  `user_id` int(11) NOT NULL,
  `date` timestamp NOT NULL DEFAULT current_timestamp(),
  `longitude` double NOT NULL,
  `latitude` double NOT NULL,
  PRIMARY KEY (`cache_id`,`user_id`),
  KEY `user_id` (`user_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci COMMENT='Custom coordinates of geocache set by user';
/*!40101 SET character_set_client = @saved_cs_client */;
```
I will update the code accordingly, however, I do not have an environment where I could test that code path

hxdimpf: Remark: I have not found out conclusively if the OCPL installation makes the `user-coords` available in the `alt_wpts` dictionary for `services/caches/geocache`. The documentation says (as you stated before):

>user-coords - extra coordinates supplied by the user who had found the cache (NOT the owner of the cache), most probably pointing to the final location of the cache (e.g. a quiz cache); this type of waypoint is available only in some installations and only if you're using Level 3 Authentication,

bekuno:
If you want a feedback from c:geo side, please open an issue there.

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,145 @@
...
+            $id = null;
+            if($row = Db::fetch_assoc($rs)) {
+                $id = $row['id'];
+            }
+            if ($id == null) {
+                Db::query("
+                    insert into cache_mod_cords (
+                        cache_id, user_id, latitude, longitude
+                    ) values (
+                        2,
```
wrygiel, line 127: Again, I see that you have a different number of values and args here. You'd really need to test this on OCPL first, before thinking of merging.

okapi/services/caches/save_user_coords/docs.xml
wrygiel: It might be that the documentation is outdated.

hxdimpf: > It might be that the documentation is outdated.

I am not exactly sure what documentation you mean. Is it this hier in `docs.xml` or somewhere else? If it is this file, the only thing I could think of is relatet to how the value (user_coords) could be retrieved in an OCPL environment? Ihave stated how it could be retrieved from OCDE but I don't actually know how id can be done in OCPL. It seems like you would know how that works in OCPL?

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,144 @@
...
+                Db::query("
+                    update cache_mod_cords
+                    set latitude  = '".Db::escape_string($latitude)."',
+                        longitude = '".Db::escape_string($longitude)."'
+                    where
+                        id = '".Db::escape_string($id)."'
+                ");
+            }
+        }
+    }
```
stefopl, line 150: OCPL ```cache_mod_cords``` table does not contain an ```id``` column.
```
create table cache_mod_cords
(
	cache_id int not null,
	user_id int not null,
	date timestamp default current_timestamp() not null,
	longitude double not null,
	latitude double not null,
	primary key (cache_id, user_id)
)
```
it should look something like this:
```
else # oc.pl branch
{
    Db::query("
        INSERT INTO cache_mod_cords (
            cache_id,
            user_id,
            latitude,
            longitude,
            date
        ) VALUES (
            '".Db::escape_string($cache_id)."',
            '".Db::escape_string($user_id)."',
            '".Db::escape_string($latitude)."',
            '".Db::escape_string($longitude)."',
            NOW()
        )
        ON DUPLICATE KEY UPDATE
            latitude = VALUES(latitude),
            longitude = VALUES(longitude),
            date = NOW()
    ");
}
```
Additionally, in OCPL only some cache types support user coordinates:
https://github.com/opencaching/opencaching-pl/blob/30ae3a4681c6840fe5aad454ccdb50fdb1340603/src/Controllers/ViewCacheController.php#L646-L648

hxdimpf: I have commited your suggested `else` path.

With respect go the cache types that support  user coordinates: in OCDE **all do**. I don't know how to access ocpl source code so pls send me the list of cache types that do support it then I will add the filtering.

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,133 @@
...
+        # Verify cache_code
+
+        $cache_code = $request->get_parameter('cache_code');
+        if ($cache_code == null)
+            throw new ParamMissing('cache_code');
+        $geocache = OkapiServiceRunner::call(
+            'services/caches/geocache',
+            new OkapiInternalRequest($request->consumer, $request->token, array(
+                'cache_code' => $cache_code,
+                'fields' => 'internal_id'
```
stefopl:
```suggestion
                'fields' => 'internal_id|type'
```

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,133 @@
...
+            ))
+        );
+        $cache_id = $geocache['internal_id'];
+
```
stefopl, line 53:
```suggestion

        self::validate_cache_type($geocache['type']);

```

okapi/services/caches/save_user_coords/WebService.php
```diff
@@ -0,0 +1,133 @@
...
+        self::update_coordinates($cache_id, $request->token->user_id, $latitude, $longitude);
+
+        $result = array(
+            'success' => true
+        );
+        return Okapi::formatted_response($request, $result);
+    }
+
```
stefopl, line 63:
```suggestion
    private static function validate_cache_type($cache_type)
    {
        if (Settings::get('OC_BRANCH') != 'oc.pl') {
            return;
        }

        $allowed_types = array(
            'Other',
            'Quiz',
            'Multi',
        );

        if (!in_array($cache_type, $allowed_types, true)) {
            throw new InvalidParam(
                'cache_code',
                "User coordinates are not supported for cache type '$cache_type'."
            );
        }
    }
```

stefopl (changes requested):
The documentation should also be updated to include these changes.

hxdimpf:
Thanks @stefopl — all suggestions applied in d27f91d:

- Fetching `type` alongside `internal_id` so the cache type is available before the DB write.
- Added `validate_cache_type()` which gates on `oc.pl` only, rejecting Traditional, Virtual, Event, Webcam, Moving, Podcast, and Own with HTTP 400 — matching OCPL's `ViewCacheController` restriction to Multi, Quiz, and Other.
- `docs.xml` updated to document the OCPL-specific type restriction.

(Personally I think OCPL should support user coordinates for all cache types, but that's a separate discussion — scoping it to the platform constraint for now.)

mlekorlz (commented):
Tested end to end on both branches with a real Level 3 OAuth token:

- OCPL: local install from the devel dump, PHP 8.1
- OCDE: local `oc-server3` (`development`) on ddev, PHP 8.2

Both at `3aab5717` with this branch applied.

The `cache_mod_cords` fix from the last round is right. The table has a composite primary key and
no `id` column, so `INSERT ... ON DUPLICATE KEY UPDATE` is the correct shape:

    CREATE TABLE `cache_mod_cords` (
      `cache_id` int(11) NOT NULL,
      `user_id` int(11) NOT NULL,
      `date` timestamp NOT NULL DEFAULT current_timestamp(),
      `longitude` double NOT NULL,
      `latitude` double NOT NULL,
      PRIMARY KEY (`cache_id`,`user_id`),
      KEY `user_id` (`user_id`)
    )

| what I ran | OCPL | OCDE |
|---|---|---|
| first save on an allowed cache type | `{"success":true}`, one row written | `{"success":true}`, one row written |
| second save, different coordinates | still one row, coords and `date` updated | still one row, updated in place (`id` unchanged, `date_created` kept, `last_modified` bumped) |
| save on a Traditional cache | HTTP 400, `"User coordinates are not supported for cache type 'Traditional'"` | accepted, as intended since the type gate is `oc.pl` only |
| read back via `services/caches/geocache` `alt_wpts` | returned as a `user-coords` waypoint with the saved coordinates | same, alongside the cache's existing parking waypoint |

Two things I checked specifically because they are branch-dependent:

`coordinates` on OCDE has `date_created` and `last_modified` as `NOT NULL` with no default, and the
insert here does not set them. That is fine: `coordinatesBeforeInsert` / `coordinatesAfterInsert`
fill them, so the insert succeeds even under `STRICT_TRANS_TABLES`, which is what the ddev install
runs.

The `type = 2` constant matches the read side. `services/caches/geocaches/WebService.php:1391-1395`
already selects `from coordinates ... and type = 2` for `oc.de`, and the round trip above confirms
the reader picks up exactly what this method writes.

No concerns from my side.

Issue #601, opened 2021-01-03
Upload field note file via OKAPI


Lineflyer:
For c:geo we recently had a [discussion](https://github.com/cgeo/cgeo/discussions/9544) about field note upload for other sites than geocaching.com.
From OC.de I know that field notes (also including duplicate handling with GC geocodes) is already implemented on their website. 

For c:geo it would just be perfect if there is a possibility to upload field notes via OKAPI.
I looked over here and other places but could not find a related issue for OKAPI. I hope this is the right place to discuss/address it.

teiling88:
Hey @Lineflyer,

can you provide me an file for this use case? I can try to implement it :-)

Greetings,
Thomas

wrygiel:
BTW, could you remind me what's the difference between field notes and [personal notes](https://opencaching.pl/okapi/services/caches/save_personal_notes.html)? (I'm not an active geocacher nowadays.)

teiling88:
Field Notes are notes about caches created with your GPS Device or an app like c:geo. So you can prepare your Log Notes and submit them finally later.

Lineflyer:
[20210105182416.txt](https://github.com/opencaching/okapi/files/5771641/20210105182416.txt)

Attached a sample field note from c:geo containing offline logs for both gc.com and oc.de caches.

hxdimpf:
I plan to implement that service method shortly.

hxdimpf:
@Lineflyer, pls review PR https://github.com/opencaching/okapi/pull/630 to see whether this would work for cgeo. The assumption is that cgep maintains one hybrid filednotes file, which may include log records for more than one platform. My PR will ignore records not starting with `OC`. Please note that the log types must case sensitively match the log types as defined for the OCDE platform. Also note that the fieldnotes file needs to have utf-8 chacacter encoding and the whole string needs to be passed as base64 encoded. Conseqently, when uploading to more than one platform during the "Export fieldnotes" dialog in cgeo, cgeo must pre-process the data before pushing.

For instance, what is a `Write note` on one platform must show up as `Comment` on OCDE. There are a few more discrepancies, even though those will be very rarely encountered.

Also, keep in mind, no UTF-16, and keep in mind the base64 encoding. And last not least: OKAPI equally supports GET and POST requests, however, since fieldnotes files can get rather large, a GET may fail due to length restrictions of URLs. My recommendation for using the service in question would be to use POST requests and put the payload into the request body.

I can make sample client code available for the cgeo team.

[...]

PR #630, opened 2023-11-25
Issue #601 fieldnotes

hxdimpf:
I am proposing the following design point:
Screenshot from 2023-12-01 10-16-51

### DIFF
--- a/okapi/core/Okapi.php
+++ b/okapi/core/Okapi.php
@@ -1128,6 +1128,30 @@ class Okapi
       * installations. That's why all known types are hardcoded here.
       * These names are officially documented and may never change!
       *
+      * These strings are the external contract: they are the values
+      * accepted in the field-notes CSV format (Garmin/Groundspeak
+      * legacy), in the services/logs/submit "logtype" parameter, and
+      * returned by services/logs/entry, services/logs/capabilities,
+      * and services/apisrv/installation. The integer IDs are an
+      * internal storage detail; API consumers never see them.
+      *
+      * | Field notes / API string  | ID | Submittable | Notes                                 |
+      * |---------------------------|----|-------------|---------------------------------------|
+      * | "Found it"                |  1 | yes         |                                       |
+      * | "Didn't find it"          |  2 | yes         |                                       |
+      * | "Comment"                 |  3 | yes         |                                       |
+      * | "Moved"                   |  4 | no          | OCPL only                             |
+      * | "Needs maintenance"       |  5 | no          | Submit via Comment + needs_maintenance2=true |
+      * | "Maintenance performed"   |  6 | no          | see issue #548                        |
+      * | "Attended"                |  7 | yes         | Event caches only                     |
+      * | "Will attend"             |  8 | yes         | Event caches only                     |
+      * | "Archived"                |  9 | yes         | OCPL: on roadmap, not on website yet  |
+      * | "Ready to search"         | 10 | yes         |                                       |
+      * | "Temporarily unavailable" | 11 | yes         |                                       |
+      * | "OC Team comment"         | 12 | no          |                                       |
+      * | "Locked"                  | 13 | no          | OCDE only                             |
+      * | "Locked, invisible"       | 14 | no          | OCDE only; maps to "Locked" in output |
+      *
       * Important: This set is not closed. Other types may be introduced
       * in the future. Developers are advised to treat unknown types as
       * 'Comment's.
--- a/okapi/core/OkapiServiceRunner.php
+++ b/okapi/core/OkapiServiceRunner.php
@@ -39,6 +39,7 @@ class OkapiServiceRunner
         'services/caches/geocaches',
         'services/caches/mark',
         'services/caches/save_personal_notes',
+        'services/caches/save_user_coords',
         'services/caches/formatters/gpx',
         'services/caches/formatters/garmin',
         'services/caches/formatters/ggz',
--- a/okapi/core/OkapiServiceRunner.php
+++ b/okapi/core/OkapiServiceRunner.php
@@ -44,6 +44,7 @@ class OkapiServiceRunner
         'services/caches/formatters/garmin',
         'services/caches/formatters/ggz',
         'services/caches/map/tile',
+        'services/draftlogs/upload_fieldnotes',
         'services/logs/capabilities',
         'services/logs/delete',
         'services/logs/edit',
--- a/okapi/services/apisrv/installation/WebService.php
+++ b/okapi/services/apisrv/installation/WebService.php
@@ -33,7 +33,38 @@ class WebService
         $result['has_image_positions'] = Settings::get('OC_BRANCH') == 'oc.de';
         $result['has_ratings'] = Settings::get('OC_BRANCH') == 'oc.pl';
         $result['geocache_passwd_max_length'] = Db::field_length('caches', 'logpw');
+        if (Settings::get('OC_BRANCH') == 'oc.de') {
+            $result['has_draft_logs'] = true;
+            $result['has_lists']      = true;
+            $result['cache_types']    = self::getCacheTypes();
+            $result['log_types']      = self::getLogTypes();
+
+        }
 
         return Okapi::formatted_response($request, $result);
+    }
+
+    private static function getCacheTypes() {
+        $rs = Db::query("
+           SELECT name
+            FROM cache_type;
+        ");
+        $cache_types = [];
+        while ($row = Db::fetch_assoc($rs)) {
+            $cache_types[] = $row['name'];
+        }
+        return $cache_types;
+    }
+
+    private static function getLogTypes() {
+        $rs = Db::query("
+           SELECT name
+            FROM log_types;
+        ");
+        $log_types = [];
+        while ($row = Db::fetch_assoc($rs)) {
+            $log_types[] = $row['name'];
+        }
+        return $log_types;
     }
 }
--- a/okapi/services/apisrv/installation/WebService.php
+++ b/okapi/services/apisrv/installation/WebService.php
@@ -33,38 +33,14 @@ class WebService
         $result['has_image_positions'] = Settings::get('OC_BRANCH') == 'oc.de';
         $result['has_ratings'] = Settings::get('OC_BRANCH') == 'oc.pl';
         $result['geocache_passwd_max_length'] = Db::field_length('caches', 'logpw');
+        $result['has_draft_logs'] = Settings::get('OC_BRANCH') == 'oc.de';
+        $result['has_lists']      = Settings::get('OC_BRANCH') == 'oc.de';
         if (Settings::get('OC_BRANCH') == 'oc.de') {
-            $result['has_draft_logs'] = true;
-            $result['has_lists']      = true;
-            $result['cache_types']    = self::get_cache_types();
-            $result['log_types']      = self::get_log_types();
-
+            $result['cache_types'] = Okapi::get_local_cachetypes();
+            $result['log_types']   = Okapi::get_submittable_logtype_names();
         }
 
         return Okapi::formatted_response($request, $result);
     }
 
-    private static function get_cache_types() {
-        $rs = Db::query("
-           SELECT name
-            FROM cache_type;
-        ");
-        $cache_types = [];
-        while ($row = Db::fetch_assoc($rs)) {
-            $cache_types[] = $row['name'];
-        }
-        return $cache_types;
-    }
-
-    private static function get_log_types() {
-        $rs = Db::query("
-           SELECT name
-            FROM log_types;
-        ");
-        $log_types = [];
-        while ($row = Db::fetch_assoc($rs)) {
-            $log_types[] = $row['name'];
-        }
-        return $log_types;
-    }
 }
--- a/okapi/services/apisrv/installation/WebService.php
+++ b/okapi/services/apisrv/installation/WebService.php
@@ -36,15 +36,15 @@ class WebService
         if (Settings::get('OC_BRANCH') == 'oc.de') {
             $result['has_draft_logs'] = true;
             $result['has_lists']      = true;
-            $result['cache_types']    = self::getCacheTypes();
-            $result['log_types']      = self::getLogTypes();
+            $result['cache_types']    = self::get_cache_types();
+            $result['log_types']      = self::get_log_types();
 
         }
 
         return Okapi::formatted_response($request, $result);
     }
 
-    private static function getCacheTypes() {
+    private static function get_cache_types() {
         $rs = Db::query("
            SELECT name
             FROM cache_type;
@@ -56,7 +56,7 @@ class WebService
         return $cache_types;
     }
 
-    private static function getLogTypes() {
+    private static function get_log_types() {
         $rs = Db::query("
            SELECT name
             FROM log_types;
--- a/okapi/services/draftlogs/upload_fieldnotes/docs.xml
+++ b/okapi/services/draftlogs/upload_fieldnotes/docs.xml
@@ -1,0 +1,66 @@
+<xml>
+    <brief>Upload Fieldnotes</brief>
+    <issue-id>630</issue-id>
+    <desc>
+        <p>This method allows you to upload a series of fieldnote objects in CSV format.
+        Fieldnote objects contain draft versions of log entries. Once uploaded, users will be able
+        to review, edit, and submit them via the Opencaching website.</p>
+    </desc>
+    <req name='field_notes' infotags='ocde-specific'>
+        <p>CSV formatted string with no header.</p>
+        <p>Each record describes a geocache draft log object consisting of four fields:</p>
+        <ol>
+            <li>Geocache Code</li>
+            <li>Date</li>
+            <li>Log Type</li>
+            <li>Log Text</li>
+        </ol>
+        <p>The first three fields are string entities that don't have line control characters in them,
+        the Log Text field is different as it may spread over muliple lines identified by line control
+        characters such as newline or linefeed and it may contain quote characters as well.
+        </p>
+        <p>The second field <i>Date</i> should be in ISO 8601 format (currently any format
+        acceptable by PHP's strtotime function also will do, but most of them don't handle
+        time zones properly, try to use ISO 8601!).</p>
+        <p>Since the log type is passed as a string, its value must match the
+        values supported by the platform (case sensitive!). In order to query
+        the names for supported log types, the service method <i>::services/apisrv/installation</i>
+        can and should be used.
+        </p>
+        <p>Note: This service method is not supported on all installations</p>
+    </req>
+    <common-format-params/>
+    <returns>
+        <p>A dictionary of the following structure:</p>
+        <ul>
+            <li>success - true</li>
+            <li>totalRecords - number of records in <i>field_notes</i></li>
+            <li>processedRecords - number of records inserted into the database</li>
+        </ul>
+        <p>processedRecords may be less than totalRecords (it may even be zero) and that
+        is the case for the following reason: Fieldnotes are created from
+        Geocaching client applications. Some of these, for instance cgeo support multiple
+        geocaching platforms from which opencaching is only one of them. Conseqently
+        Fieldnotes may be  a "hybrid object" which may ontain records targeted at more than one
+        platform. For instance for geocaching.com logs, the records start with <b>GC....</b>
+        while on opencaching.de the log records start with <b>OC....</b>. Other opencaching
+        platforms use other codes, for instance opencaching.pl uses <b>OP...</b>.</p>
+        <p>
+        The client application may upload one and the same Fielnotes object to all platforms and
+        it is within the platform's discretion
+        to filter out what matches their object definition.
+        opencaching.de discards everything that doesn't start with "OC."</p>
+        <p>In addition, in that hybrid object there will be <i>Log Type</i>, a string that
+        inevitably has a different definition for different platforms. For instance, what
+        is called a "Write note" log on geocaching.com is recognized as "Note" some
+        opencaching platforms or "Comment" on others.
+        Consequently fieldnotes records may have Log Types which are not understood
+        by the target platform in which case they will be discarded without notice.</p>
+        <p>It is the responsibility of the client application to assign the correct Log Type
+        string when the offline log is created.
+        Sending log type names which are not supported by the designated target platform
+        is considered a programming error within the client application. In order
+        to determine a target's supported log type names the service: <i>::services/apisrv/installation</i>
+        can and should be used.</p>
+    </returns>
+</xml>
--- a/okapi/services/draftlogs/upload_fieldnotes/docs.xml
+++ b/okapi/services/draftlogs/upload_fieldnotes/docs.xml
@@ -1,5 +1,6 @@
-<xml infotags='ocde-specific'>
+<xml>
     <brief>Upload Fieldnotes</brief>
+    <infotags>ocde-specific</infotags>
     <issue-id>630</issue-id>
     <desc>
         <p>This method allows you to upload a series of fieldnote objects in CSV format.
--- a/okapi/services/draftlogs/upload_fieldnotes/docs.xml
+++ b/okapi/services/draftlogs/upload_fieldnotes/docs.xml
@@ -1,66 +1,46 @@
-<xml>
+<xml infotags='ocde-specific'>
     <brief>Upload Fieldnotes</brief>
     <issue-id>630</issue-id>
     <desc>
         <p>This method allows you to upload a series of fieldnote objects in CSV format.
-        Fieldnote objects contain draft versions of log entries. Once uploaded, users will be able
+        Fieldnotes are draft versions of log entries. Once uploaded, users will be able
         to review, edit, and submit them via the Opencaching website.</p>
+        <p>This method is only available on installations that support draft logs. Check
+        the <b>has_draft_logs</b> field in
+        <a href="%OKAPI:methodargref:services/apisrv/installation%">services/apisrv/installation</a>
+        before calling it.</p>
     </desc>
-    <req name='field_notes' infotags='ocde-specific'>
-        <p>CSV formatted string with no header.</p>
-        <p>Each record describes a geocache draft log object consisting of four fields:</p>
+    <req name='field_notes'>
+        <p>The fieldnotes data. Accepted either as a raw string or base64-encoded
+        (base64 is recommended when the log text contains newlines or special characters,
+        and is required when using the OKAPI Browser).</p>
+        <p>The data is a sequence of CSV records with <b>no header line</b>. Each record
+        has four fields:</p>
         <ol>
-            <li>Geocache Code</li>
-            <li>Date</li>
-            <li>Log Type</li>
-            <li>Log Text</li>
+            <li><b>Geocache Code</b> - e.g. <tt>OC1012</tt></li>
+            <li><b>Date</b> - ISO 8601 UTC, e.g. <tt>2023-11-23T10:17:55Z</tt></li>
+            <li><b>Log Type</b> - must match a value from the <b>log_types</b> list in
+            <a href="%OKAPI:methodargref:services/apisrv/installation%">services/apisrv/installation</a>
+            (case sensitive). Records with unrecognized log types are silently skipped.</li>
+            <li><b>Log Text</b> - may span multiple lines and contain quote characters.</li>
         </ol>
-        <p>The first three fields are string entities that don't have line control characters in them,
-        the Log Text field is different as it may spread over muliple lines identified by line control
-        characters such as newline or linefeed and it may contain quote characters as well.
-        </p>
-        <p>The second field <i>Date</i> should be in ISO 8601 format (currently any format
-        acceptable by PHP's strtotime function also will do, but most of them don't handle
-        time zones properly, try to use ISO 8601!).</p>
-        <p>Since the log type is passed as a string, its value must match the
-        values supported by the platform (case sensitive!). In order to query
-        the names for supported log types, the service method <i>::services/apisrv/installation</i>
-        can and should be used.
-        </p>
-        <p>Note: This service method is not supported on all installations</p>
+        <p>Character encoding is detected automatically. Supported encodings: UTF-8
+        (with or without BOM), UTF-16LE (with or without BOM), UTF-16BE (with BOM).
+        This covers all known fieldnote sources (cgeo, Garmin, Looking4Cache, etc.).</p>
+        <p>Fieldnotes produced by multi-platform apps (e.g. cgeo) typically contain
+        records for several geocaching platforms at once. Records not starting with
+        an OC cache code are silently skipped; the difference between
+        <b>total_records</b> and <b>processed_records</b> in the response reflects this.</p>
     </req>
     <common-format-params/>
     <returns>
         <p>A dictionary of the following structure:</p>
         <ul>
-            <li>success - true</li>
-            <li>totalRecords - number of records in <i>field_notes</i></li>
-            <li>processedRecords - number of records inserted into the database</li>
+            <li><b>success</b> - true</li>
+            <li><b>total_records</b> - number of CSV records found in the input</li>
+            <li><b>processed_records</b> - number of records actually inserted as draft
+            logs; may be less than <b>total_records</b> when records are skipped (wrong
+            platform prefix, unsupported log type, unknown cache code, invalid date)</li>
         </ul>
-        <p>processedRecords may be less than totalRecords (it may even be zero) and that
-        is the case for the following reason: Fieldnotes are created from
-        Geocaching client applications. Some of these, for instance cgeo support multiple
-        geocaching platforms from which opencaching is only one of them. Conseqently
-        Fieldnotes may be  a "hybrid object" which may ontain records targeted at more than one
-        platform. For instance for geocaching.com logs, the records start with <b>GC....</b>
-        while on opencaching.de the log records start with <b>OC....</b>. Other opencaching
-        platforms use other codes, for instance opencaching.pl uses <b>OP...</b>.</p>
-        <p>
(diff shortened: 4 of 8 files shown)


hxdimpf:
Just noticed that this PR now contains **both** feature branches I have created recently. I don't know how to focus on each of the branches separately with git. I am open to suggestions with respect to rectifying the situation.

hxdimpf:
@wrygiel: could you please review this pull request when you have a chance? Thank you!

wrygiel:
Hi! I'm not much familiar with this. Can you point me to some docs explaining what are field notes, and how they differs from log entries?

hxdimpf:
> Hi! I'm not much familiar with this. Can you point me to some docs explaining what are field notes, and how they differs from log entries?

I am not aware of any documentation but I can quickly explain what it is and also post an example. Some applications on the smartphone, such as: cgeo, Locus Map, Garmin, Looking4Cache, Cachly provide a capability to "offline log" a geocache out in the field, an important function if you don't have a mobile carrier out there. These offline logs end up in a file (or internal data structure) that consists of one or more CSV formatted records like this:

```
OC1006,2023-11-23T10:17:55Z,Found it,"lorem ipsum ipsum lorem"
OC1013,2023-11-23T10:17:46Z,Found it,"Thx to Chaos"
```

There are some complications associated with this though. First: It is not standardized. Some tools encode `UTF-16`, either `LE` (little endian) or `BE` (big endian) and there frequently is a `BOM` (Byte Order Marker), which some tool include and others don't.

For this reason I explilcitly require the client app to convert it on their side to character encoding `utf-8` and in addition pack the whole string into `base64`. The latter is required as the  last field in the CSV, the draft log text, may actually span multiple lines in the file and it may contain ASCII control characters such as line feeds, carriate returns, and single and double quotes.

The OCDE website can parse such a file via upload from the browser. The user can then, for each draft log, edit the log text and finally submit it a real log.

With this PR I provide the capability to client applications to directly upload their fieldnotes file into the database as draft logs. Once they are up, there is no difference compared to manual upload via the browser and the filesystem.

wrygiel:
Isn't it the same as submitting a series of logs via `services/logs/submit`?

Is the idea to move some work related to processing the file from the app to the server?

hxdimpf:
> Isn't it the same as submitting a series of logs via `services/logs/submit`?
> 
> Is the idea to move some work related to processing the file from the app to the server?

I apologize for being not precise enough: It is not the same as submitting log. Submitting a log is the final transaction that also records a statistics update for the cacher and more importantly makes this log visible in the cache listing.

The upload fieldnotes function puts these logs into a draft state, they will not be commited as real logs yet. There go into a separate table in the database (details). There is a UI in OCDE (as it is on other platforms) to manage these draft logs, e.g. by  editing the log text, change time/date, perhaps changinf the log type and then, at the end, press the "submit" button which finally commits a draft log into a real log and this making it visible to othersin the cache listing.

Does this help? Let me know if it doesn't

wrygiel:
Ah, got it, so a "field note" is a "draft log entry". Yes, I can review it now.

okapi/services/fieldnotes/upload/docs.xml
```diff
@@ -0,0 +1,49 @@
+<xml>
```
wrygiel, line 1: Make sure that your method returns the `ocde-specific` infotag in [`services/apiref/method`](https://www.opencaching.de/okapi/services/apiref/method.html).

okapi/services/fieldnotes/upload/docs.xml
```diff
@@ -0,0 +1,49 @@
...
+    <brief>Upload Fieldnotes</brief>
+    <issue-id>630</issue-id>
+    <desc>
+        <p>Upload a set of one or more fieldnotes records.</p>
```
wrygiel, line 5: I think we can be more detailed here. Many developers (particularly the ones from countries running the OCPL branch) will not know what "fieldnotes" are. How about this:

This method allows you to upload a series of fieldnotes in the CSV format. Fieldnotes are draft versions of log entries. Once uploaded, users will be able to review, edit, and submit them via the Opencaching site.

okapi/services/fieldnotes/upload/docs.xml
```diff
@@ -0,0 +1,49 @@
...
+    </desc>
+    <req name='field_notes'>
+        <p>Fieldnotes consist of one or more records in CSV format. Each
+        record comprises a geocache log consisting of four fields:</p>
```
wrygiel, line 9: CSV format specification mentions an optional header. Should we specify if this header is required, optional, or forbidden here? (It would be much easier if there was some external specification for the "fieldnotes file", but you mentioned there is no such specification, so we may need to specify all such details ourselves.)

okapi/services/fieldnotes/upload/docs.xml
```diff
@@ -0,0 +1,49 @@
...
+        <ul>
+            <li>Geocache Code</li>
+            <li>Date</li>
+            <li>Log Type</li>
+            <li>Log Text</li>
+        </ul>
+        <p>The first three fields are simple entities, the Log Text field is
```
wrygiel, line 16: Can you clarify what you mean by "simple entities"?

I suspect "Geocache Code" and "Log Types" are plaintext strings, but the "Date" needs more clarification. There are many ways to format a date. Is it ISO 8601?

okapi/core/OkapiServiceRunner.php
```diff
@@ -42,6 +42,7 @@ class OkapiServiceRunner
         'services/caches/formatters/gpx',
         'services/caches/formatters/garmin',
         'services/caches/formatters/ggz',
+        'services/fieldnotes/upload',
```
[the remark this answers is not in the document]
wrygiel: I wonder if it wouldn't be easier to add a new parameter to the `submit` method instead - which, if given, would insert the log entry to the "draft" table, instead of the final table. This would allow the devs to use an existing method which they're already familiar to. On the other hand, it would not support the "fieldnotes" format. Are you confident devs prefer to this format over JSON?

okapi/services/fieldnotes/upload/docs.xml
```diff
@@ -0,0 +1,49 @@
...
+        <p>Note: This service method is not supported on all installations</p>
+    </req>
+    <common-format-params/>
+    <returns>
+        <p>A dictionary of the following structure:</p>
+        <ul>
+            <li>success - true</li>
+            <li>totalRecords - number of records in <i>field_notes</i></li>
+            <li>processedRecords - number of records inserted into the database</li>
+        </ul>
```
wrygiel, line 47: This implies that some records might not be inserted. That makes sense, because it's always possible for the caller to give us invalid entries which cannot be inserted. However, we should clarify when and why this may happen. Also, success should probably not be true in this case?

wrygiel (commented):
Reviewed the docs only for now. Once we agree on the API, I will review the code too.

okapi/services/fieldnotes/upload/docs.xml
hxdimpf: There will be records in that dataset that we need to skip. Cachers using those tools, e.g. cgeo, do caching on geocaching.com as well as opencaching.de and perhaps even more platforms all at the same time. The resulting fieldnotes dataset will have records that start with "GC....,....,..." and "OC.....,....,....". The PR filters out everything that is irrelevant for OCDE and will inform the user in the result via those counters.

Even if the record starts with "OC" it may still be discarded if the Log Type is not known to OCDE. Unfortunately there is no standardization across platforms and giving an ASCII text as log type is a really bad choice but it is the way it is and we can't do anything about it except discarding what doesn't fit into the platform.

okapi/services/fieldnotes/upload/docs.xml
hxdimpf: Simple entities are strings withouit line control characters in it. The date is in the format:
```
2023-11-23T10:17:55Z
```
Don't know at this time what this format is called. PR handles it appropriately, I believe

okapi/services/fieldnotes/upload/docs.xml
wrygiel: Thanks! Can you clarify all that in the file?

okapi/services/fieldnotes/upload/docs.xml
wrygiel: This looks like ISO 8601. Do you always expect the timezone to be UTC?

hxdimpf: I have seen data from: cgeo, Garmin, Looking4Cache, they all com in zulu time

wrygiel: Then you're looking for something like:

The Date field needs to be formatted in a very specific subset of the ISO 8601 format - it needs to be a "2023-11-23T10:17:55Z" string, representing the date and time in UTC timezone.

What will happen when the date and/or cache code is invalid - will the entry be ignored, or will the error be raised?

okapi/services/fieldnotes/upload/docs.xml
hxdimpf: done

okapi/services/fieldnotes/upload/docs.xml
hxdimpf: done

okapi/core/OkapiServiceRunner.php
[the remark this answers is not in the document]
hxdimpf: The first part, with respect to selecting a better namespace is ok with me.

The second, merging it into existint log methods is not, at least not yet. I do not yet have sufficient insight into OKAPI in order to have a position here. I can only assume, when you are making such a suggestion that merging it into the submit method would save us from duplicating functionality. Functionality that is already available in that method and we would simply reuse it? You are further assuming that adding if then else conditions into the existing function and targetting different database tables is the lesser evil? One major motivation for me to not touch existing methods is to guarantee that I will not break existing functions. Also I am quite sure there are client applications out there consuming OKAPI, that are no longer actively maintained and it may be a show stopper for those if all of a suden a method behaves only slightly different than it did in the past. It is a problem even it is actually a bug in OKAPI as there is nobody out there chasing it down to OKAPI, some appications will just silently no longer work.

I'd say: look at the PR's code and find out whether the logic is complex and could be simplified by moving stuff into the submit method. I'd clearly need help in doing so as those functions aren't really simple. But that should not keep us from not doing it, if it makes sense for other reasons.

I can only describe what I had in mind with those fieldrecords in general and let us clean up the name space a bit here:

`draftlogs/upload - we have discussed this`
`draftlogs/query  - return a list of ids to later target the methods below`
`draftlogs/flush - remove all draft logs from the database`
`draftlogs/delete - delete a specific draft log based on what query returned`

Also keep in mind that the `upload` operates on potentially very large arrays. It is not uncommon that such fieldnotes files have hunderts of entries. Throwing a particular record into the right database isn't the challenge here. The challenge lies in parsing the input data appropriaetely. Well, while I am writing this, is your thinking perhaps to leave ll the parsing as it is and at the point we deal with the database we'd call an internal service such as this one:
```
$geocache = OkapiServiceRunner::call(
            'services/caches/geocache',
            new OkapiInternalRequest($request->consumer, $request->token, array(
                'cache_code' => $cache_code,
                'fields' => 'my_notes|internal_id'
            ))
        );

```
... of course for `logs/submit` then?

okapi/services/fieldnotes/upload/docs.xml
hxdimpf: > Then you're looking for something like:
> 
> The Date field needs to be formatted in a very specific subset of the ISO 8601 format - it needs to be a "2023-11-23T10:17:55Z" string, representing the date and time in UTC timezone.

I had actually assumed and still do that this handles date/time appropriately regardles whether it is UTC oder not:

`$date  = date("Y-m-d H:i:s", strtotime($n['date']));`

Of course, if the resulting `$date` variable is not what we expect (how would we test anything other than perhaps null?), then I had intended to discard the record, however, I just notices I haven't handled that condition yet in the PR's code.

> 
> What will happen when the date and/or cache code is invalid - will the entry be ignored, or will the error be raised?

I would opt for discarding the record and not throwing an exception. But an exception would be fine with me too.

okapi/core/OkapiServiceRunner.php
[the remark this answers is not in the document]
wrygiel: I suggested that because I suspect it would be easier for external developers, API-wise. Imagine an app which already is able to submit log entries, and imagine the developer wants to add a "save as draft instead" checkbox to this app. In such case, adding an argument to the existing `submit` method would be a simple way of providing them with such functionality.

It does have many drawbacks however - e.g. it would obviously be more difficult to implement on our side, and we some arguments of the `submit` method would not work (especially given that OCDE keeps draftlogs and regular logs in separate tables). I also agree that you probably should not touch any existing methods until you're able to test them on OCPL database.

So to sum up, I agree it's good to keep it separate. +1 for creating the `draftlogs` directly, but let's stick with existing naming conventions for the methods (e.g. we don't have `query` methods, but we have `geocache` and `geocaches`), so this would be:

```
services/draftlogs/upload_fieldnotes_file
services/draftlogs/draftlog (retrieves one)
services/draftlogs/draftlogs (retrieves many)
services/draftlogs/add (JSON alternative to use instead upload_fieldnotes_file for better error control etc.)
services/draftlogs/delete
```

okapi/services/fieldnotes/upload/docs.xml
wrygiel: I suggest to copy the docs and implementation from other methods which already take the date argument in ISO 8601 format. See `when` argument [here](https://www.opencaching.de/okapi/services/logs/submit.html).

okapi/core/OkapiServiceRunner.php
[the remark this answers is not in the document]
hxdimpf: All good what you're suggesting. I am also with you with respect to the JSON except that no single client application exists that would use it and just like so many other applications most of these are virtually dead. cgeo is the only one that really lives. We need to discuss this with @Lineflyer whether thinks cgeo would want to use the JSON path,

I will update the PR accordingly. e.g. moving stuff into the right namespace and I will leave it with just the upload for now as cgeo needs the upload. They would never use any of the other methods.

okapi/services/draftlogs/upload_fieldnotes_file/WebService.php
```diff
@@ -0,0 +1,175 @@
...
+
+        $lines = explode("\n", $fieldnotes);
+        $lines = array_filter($lines); // Drop empty lines
+
+        foreach ($lines as $line) {
+            if ($start) {
+                $buffer = $line;
+                $start = false;
+            } else {
+                if (strpos($line, 'OC') !== 0) {
```
wrygiel, line 131: I don't understand how newlines are handled in this file. How can you recognize "newline which is part of the log entry" from "newline which separates two CSV rows"?

```
OC1234,Date,Found it,This first line of multiline log.
OC is awesome! That's the second line of a multiline log.
```

okapi/services/fieldnotes/upload/docs.xml
```diff
@@ -0,0 +1,69 @@
...
+        <p>processedRecords may be less than totalRecords (it may even be zero) and that
+        is the case for the following reason: Fieldnotes files are created from
+        Geocaching client applications on Smartphones. These applications support multiple
+        geocaching platforms from which opencaching is only one of them. Conseqently the
+        Fieldnotes file is a "hybrid file" which contains records for multiple different
+        platforms. For instance for geocaching.com logs the records start with <b>GC....</b>
+        while of opencaching the log records start with <b>OC....</b>. The client application
+        uploads one and the same Fielnotes file to all platforms and it is the platform's
+        task to filter out what matches their objects. Geocaching.com will discard "OC logs"
+        and opencaching must discard "GC logs" in that file.</p>
```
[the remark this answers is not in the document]
hxdimpf: Do other opencaching platforms use "OCxxx" codes? If yes, how can I differentiate an OCPL log from an OCDE log?

okapi/services/fieldnotes/upload/WebService.php
```diff
@@ -0,0 +1,175 @@
...
+        if (!$start) {
+            $output[] = trim($buffer);
+        }
+
+        return $output;
+    }
+
+    // ------------------------------------------------------------------
+
+    private static function CSVtoArray($text)
```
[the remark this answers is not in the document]
hxdimpf: Of courfse I have tied to use that function and believe me or not, I have spent hours to parse such a string. Ultimately and years ago I wrote those function in JS and they are proven to work so I converted it to php and used it. The problem is in the structure of the CSV record. A CSV record is not a 1:1 match with a "line in a file". The last field of the record, which is a draft log, can be anything the user has composed on the client application. It typicall streches over multiple lines in the fila, it hast control characters in it such as line feeds and carriage returns, it has also single quotes in it, matched or unmached paris, it has double quotes in it, believe me, such trivial functions like `str_getcsv()` miserably fail.

Here is a sampe, however, pls note that the full implications can only be understood if I'd send you a real file. The following file has two records, each 4 fields. The fourth field is tough:

```
OC1012,2023-11-23T10:17:55Z,Found it,"lorem ipsum ipsum lorem

bla
bla

Thx to gcnuf  for 'Brunnen 1: Blick auf Herrenberg von oben',
JoeDante

Take only what you need and be on your way ..."
OC100B,2023-11-23T10:17:46Z,Found it,"Thx to Chaos truppe  for 'cachespätzle',
JoeDante

Take only what you need and be on your way ..."
```

okapi/services/fieldnotes/upload/WebService.php
```diff
@@ -0,0 +1,175 @@
...
+                    )
+                ");
+
+            }
+            $result = array(
+                'success'          => true,
+                'totalRecords'     => $notes['totalRecords'],
+                'processedRecords' => $notes['processedRecords']
+            );
+            //$result = json_encode($notes, JSON_PRETTY_PRINT); // debug
```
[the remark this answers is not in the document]
hxdimpf: removed

okapi/services/draftlogs/upload_fieldnotes_file/WebService.php
hxdimpf: See my detailed explanation above regarding the format and rationales for using my own custom functions.

BTW: I also spent hours passing the file, which is actually a long string, directly as utf-8 string in the POST body. I have not accomplished to generate a Oauth Level 3 signature ever. Once I base64 encoded it, the problems went away instantly. I understand that all you had back then was Oauth1 but I understand even much much more why Oauth2 was invented.

okapi/services/fieldnotes/upload/WebService.php
```diff
@@ -0,0 +1,175 @@
...
+                $geocache = OkapiServiceRunner::call(
+                'services/caches/geocache',
+                    new OkapiInternalRequest($request->consumer, $request->token, array(
+                        'cache_code' => $n['code'],
+                        'fields' => 'internal_id'
+                    ))
+                );
+                $user_id     = $request->token->user_id;
+                $geocache_id = $geocache['internal_id'];
+                $type        = Okapi::logtypename2id($n['type']);
```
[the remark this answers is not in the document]
hxdimpf: They will never be really committed, just end up as drafts

okapi/services/fieldnotes/upload/WebService.php
```diff
@@ -0,0 +1,175 @@
...
+        );
+
+        if (Settings::get('OC_BRANCH') == 'oc.de')
+        {
+
+            $field_notes = $request->get_parameter('field_notes');
+            if (!$field_notes) throw new ParamMissing('field_notes');
+
+            $notes = self::parse_notes($field_notes);
+            if ($notes['success'] === false) throw new InvalidParam('field_notes', "Input data not recognized.");
```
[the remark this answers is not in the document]
hxdimpf: I understand the point. Obviously you have been confronted with this type of complaints.

Do you have a suggestion how this could be made better?  Since we're throging an exception, I can handle it deeper down and not even let it bubble up to this point. .... Let me make an attempt ....

okapi/services/fieldnotes/upload/docs.xml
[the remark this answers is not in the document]
wrygiel: Other OKAPI installations use other prefixes (e.g. `OP` for OCPL). What I mean in the comment above is that, in your comment, you probably should also mention that OC logs from other installations will also be discarded (as opposed to only "GC logs", as you currently document it).

okapi/services/fieldnotes/upload/WebService.php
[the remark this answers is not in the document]
wrygiel: IMO best practice is to throw and catch exceptions.

okapi/services/draftlogs/upload_fieldnotes_file/WebService.php
wrygiel: This seems to be a problem with your OAuth signing library then? There are many other methods which take UTF-8 strings and they do not require Base64.

BTW, as you mentioned before, you can use `PLAINTEXT` authentication with https, so you do not need to sign your requests anymore.

okapi/services/fieldnotes/upload/WebService.php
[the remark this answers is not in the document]
wrygiel: By "committed" I mean "stored in the database".

okapi/services/draftlogs/upload_fieldnotes_file/WebService.php
hxdimpf: > This seems to be a problem with your OAuth signing library then? There are many other methods which take UTF-8 strings and they do not require Base64.

I am not using any library, because I couldn't get any to work in node.js. If you're really curious I can send you my testing code, it is just a few lines of JS then you can perhaps see that my signature calculation is correct and still it dies not work if the non base64 encoded string is in there.
 
> BTW, as you mentioned before, you can use `PLAINTEXT` authentication with https, so you do not need to sign your requests anymore.

The OCDE docker based dev env runs only over http. PLAINTEXT is not an option for my testing, unfortunately.

BTW: http or https, PLAINTEXT or HMAC: If I try to put a multiline string into the OKAPI Brower's entry field, how is that supposed to work? No problem with a base64 string.

okapi/services/fieldnotes/upload/docs.xml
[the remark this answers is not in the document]
hxdimpf: Ah ok, didn't know they have their own prefix, will clarify that tomorrow in my next commit then

okapi/services/fieldnotes/upload/WebService.php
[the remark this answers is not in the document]
wrygiel: Ah, no, there's something off. "OC100B" gets glued to the previous line.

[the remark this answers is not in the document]
hxdimpf: Interesting! This looks actually the way it should. I need to investigate, where in tha path the string gets screwed up then, perhaps in the various `encodeURIComponent()` calls. The reason why base64 works is most likely because at no point anything is messing with the string.

More and more I come to the conclusion that we should require the client application to build a clean JSON array which can be stringified at the sending end and parsed at the receiving end.

At this time it would be `cgeo` which would have to adapt. I am not aware of any other client application uploading fieldnotes. Perhaps Garmin can upload directly to geocaching.com, that I don't know for sure but I doubt it.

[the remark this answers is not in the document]
wrygiel: [Here's a fixed version](https://www.tehplayground.com/L0BxcjwFRDm9stmy). I somehow assumed that `str_getcsv` reads the entire CSV, whereas it reads only one row, so this needs to be done in a loop.

okapi/services/draftlogs/upload_fieldnotes_file/WebService.php
wrygiel: You can hack it locally, so that it thinks that it's https. See `OkapiOAuthServer.php` line 16.

I think OKAPI Browser doesn't support multiple, sadly. I won't be able to release a new version.

hxdimpf: > You can hack it locally, so that it thinks that it's https. See `OkapiOAuthServer.php` line 16.

Good pointer! Thx much for your help, not just here, but in general. I treally appreciate it a lot.

First I will revisit the CSV parsing on the server side and once that really works without my custom function I will try to send PLAINTEXT signed non-base64 data and see how it goes.

> 
> I think OKAPI Browser doesn't support multiple, sadly. I won't be able to release a new version.

No problem. It is of great help the way it is. For development I use the base64. Should we figure out at the end, how to do without it, it is a two line change to remove the base64 decoding step.

wrygiel: Happy to help!

hxdimpf:
A remark on character encoding of the `field_notes` parameter: As I already stated, there is no documentation of the format. The layout of the CSV columns is clear, however the character encoding is not. Different applications use different encodings. We could take the approach and say client applications send whatever character encoding they use internally and OKAPI will be smart enough to make us of it. This complicates OKAPI of course. The PR in the current state took the approach: Client, you convert to utf-8 encoding and all is fine. I know that this will cause burden to cgeo as cgeo maintains a fieldnotes file which it typically sends via an API to geocaching.com. Ideally cgeo would want to send the same data over OKAPI, least amount of effort on cgeo's side.

However, and for whatever reason cgeo uses an obscure encoding, which I believe is not even standardized. I also know that cgeo hast changed that a few years to what it is now. For reference I include a cgeo fieldnotes file header in hex such that you can draw your own conclusion:

`00000000: 4f00 4300 4400 3300 4100 4400 2c00 3200  O.C.D.3.A.D.,.2.`

The only thing that's clear is that it is a 16-bit encoding, however:

```
enca -L none 20231127092821.txt 
Universal character set 2 bytes; UCS-2; BMP
  LF line terminators
  Byte order reversed in pairs (1,2 -> 2,1)
```

It seems to be UTF-16LE without BOM.

So what do we do, accept this format? Tell cgeo to encode in UTF-8? Be very smart and accept it all whatever it is?

SammysHP:
> However, and for whatever reason cgeo uses an obscure encoding, which I believe is not even standardized.

The uncommon encoding has historic reasons. It was specified/implemented that way a long, long time ago together with Garmin GPS receivers and hasn't changed since then.

okapi/core/OkapiServiceRunner.php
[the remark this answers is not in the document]
Lineflyer: @hxdimpf I am probably the wrong person to discuss "coding stuff" as I am no programmer myself but only a bit responsible for support and issue sorting and alike in c:geo.
Best is to ask your question or your request to c:geo directly in a matching issue on the c:geo Github page.

hxdimpf:
Addressed all remaining open items in 29bdd99:

**Code fixes:**
- Geocache lookup now wrapped in try/catch → unknown/missing cache codes (GC, OP, or non-existent OC codes) skip the record instead of crashing the request. This is the correct behavior for a batch operation.
- Invalid dates now skip the record instead of throwing `InvalidParam`.
- `processed_records` is now counted after each successful DB insert — previously it was counting type-filtered records in `parse_notes`, which could overstate the actual number of inserts.
- Removed the debug method and the commented-out debug call.
- Fixed `has_draftlogs` typo → `has_draft_logs` in the `BadRequest` message.

**Docs:**
- `infotags='ocde-specific'` moved from the `<req>` element to `<xml>` — the whole method is OCDE-only, not just one parameter.
- Response field names corrected: `totalRecords`/`processedRecords` → `total_records`/`processed_records` (matching what the PHP actually returns).
- `field_notes` parameter docs now describe: base64 input support, automatic encoding detection (UTF-8 with/without BOM, UTF-16LE with/without BOM, UTF-16BE), skip reasons, and a reference to `services/apisrv/installation` for supported log type names.
- `services/apisrv/installation` docs now document the `has_draft_logs` field (it was already returned by the code but never documented in the API reference).

Re `str_getcsv`: it is used inside `parse_notes` on each fully-assembled record string — but only *after* the encoding detection and multi-line reassembly are done by the custom `parse_csv` layer. The encoding mess (UTF-16LE without BOM is the default for Garmin/cgeo) makes it impossible to skip that layer.

stefopl:
> * `infotags='ocde-specific'` moved from the `<req>` element to `<xml>` — the whole method is OCDE-only, not just one parameter.

Was anything else besides moving `infotags` verified to ensure it works as expected?
The methods `/okapi/services/apiref/method` and `/okapi/services/apiref/method_index` should correctly return `infotags`.

okapi/services/draftlogs/upload_fieldnotes/docs.xml
```diff
@@ -0,0 +1,46 @@
+<xml infotags='ocde-specific'>
+    <brief>Upload Fieldnotes</brief>
+    <issue-id>630</issue-id>
```
stefopl:
```suggestion
<xml>
    <brief>Upload Fieldnotes</brief>
    <infotags>ocde-specific</infotags>
    <issue-id>630</issue-id>
```

stefopl: Please fix all <xml infotags='ocde-specific'> entries in both PRs.

okapi/services/apisrv/installation/WebService.php
```diff
@@ -33,7 +33,38 @@ public static function call(OkapiRequest $request)
         $result['has_image_positions'] = Settings::get('OC_BRANCH') == 'oc.de';
         $result['has_ratings'] = Settings::get('OC_BRANCH') == 'oc.pl';
         $result['geocache_passwd_max_length'] = Db::field_length('caches', 'logpw');
+        if (Settings::get('OC_BRANCH') == 'oc.de') {
+            $result['has_draft_logs'] = true;
+            $result['has_lists']      = true;
+            $result['cache_types']    = self::get_cache_types();
+            $result['log_types']      = self::get_log_types();
```
stefopl: This is missing from the documentation.

okapi/services/apisrv/installation/WebService.php
```diff
@@ -33,7 +33,38 @@ public static function call(OkapiRequest $request)
         $result['has_image_positions'] = Settings::get('OC_BRANCH') == 'oc.de';
         $result['has_ratings'] = Settings::get('OC_BRANCH') == 'oc.pl';
         $result['geocache_passwd_max_length'] = Db::field_length('caches', 'logpw');
+        if (Settings::get('OC_BRANCH') == 'oc.de') {
+            $result['has_draft_logs'] = true;
+            $result['has_lists']      = true;
```
stefopl:
```suggestion
        $result['has_draft_logs'] = Settings::get('OC_BRANCH') == 'oc.de';
        $result['has_lists']      = Settings::get('OC_BRANCH') == 'oc.de';
        if (Settings::get('OC_BRANCH') == 'oc.de') {
```
I'm not fully convinced that conditionally omitting these fields is the best approach.

The docs say "Only present on installations where it is true", but from a practical API consistency point of view it might be better to always return these fields and use false for installations where the feature is not supported. That would make the response structure stable and easier to consume on the client side.

stefopl: ```$result['has_lists']``` is currently missing from documentation.

I see it is already documented in the next PR, but since this value is already being returned here, it should probably be documented in this PR as well. Otherwise, if only this PR gets merged, the API response will include a field that is not described anywhere.

okapi/services/apisrv/installation/WebService.php
stefopl: Why wasn’t it possible to use ```Okapi::get_local_cachetypes()``` and ```Okapi::get_submittable_logtype_names()``` here instead?

stefopl (changes requested):

hxdimpf:
Addressed all items in 1cb7fe6:

**infotags format (`docs.xml`)**
Changed `<xml infotags='ocde-specific'>` to `<xml>` with `<infotags>ocde-specific</infotags>` as a child element. The `apiref/method` service reads `$docs->infotags` (a child element), not an attribute on `<xml>`, so the attribute form was silently ignored. Now the method will correctly appear as OCDE-specific in the method index and method detail responses.

**Always return `has_draft_logs` and `has_lists`**
Accepted the suggestion. Both fields are now set unconditionally as `Settings::get('OC_BRANCH') == 'oc.de'`, returning `false` on non-OCDE installations instead of omitting the key. `cache_types` and `log_types` remain inside the `if` block since they are only meaningful where the feature exists.

**Use `Okapi::get_local_cachetypes()` / `Okapi::get_submittable_logtype_names()`**
Replaced the two private DB-query methods with these existing Okapi helpers and removed the now-unused private methods. This ensures the values returned by `installation` are the same canonical names that `upload_fieldnotes` uses internally (`logtypename2id` / `get_submittable_logtype_names`), eliminating any possible mismatch between advertised and accepted values.

**`has_lists` and missing fields in docs**
Added documentation for `has_lists`, `cache_types`, and `log_types` in `installation/docs.xml`. The `cache_types` and `log_types` fields were already being returned by the code but were never documented. Also updated the `has_draft_logs` description to drop "Only present on installations where it is true" since it is now always present.

okapi/core/Okapi.php
```diff
@@ -1128,6 +1128,30 @@ public static function cache_size2_to_oxsize($size2)
...
+      * | "Archived"                |  9 | yes         | OCPL: on roadmap, not on website yet  |
+      * | "Ready to search"         | 10 | yes         |                                       |
+      * | "Temporarily unavailable" | 11 | yes         |                                       |
+      * | "OC Team comment"         | 12 | no          |                                       |
+      * | "Locked"                  | 13 | no          | OCDE only                             |
+      * | "Locked, invisible"       | 14 | no          | OCDE only; maps to "Locked" in output |
+      *
       * Important: This set is not closed. Other types may be introduced
       * in the future. Developers are advised to treat unknown types as
       * 'Comment's.
```
andrixnet, line 1157: Maybe include in the code a reference to https://wiki.opencaching.eu/index.php?title=Log_types as additional documentation?

PR #631, opened 2023-11-28
issue#627 listservices: Full suite of managing geocache lists via OKAPI services

hxdimpf:
This PR addresses the full suite of list related functions in OKAPI. At this time I will put it into draft state such that the semantics can be reviewed and possibly adjusted accordingly. Let me give some rationalization for the design pattern used:

**Description:**

In this design pattern, we have adopted a modular and purpose-driven approach to handle various aspects of list management within our webservices. The separation of functionality into distinct services aims to enhance maintainability, extensibility, and clarity in code structure.

**1. ** ***Create, Update, Delete (CUD) Services:*** The CUD services (Create, Update, Delete) focus on the fundamental operations related to the creation, modification, and removal of lists. Each service is tailored to handle specific aspects of list metadata, such as name, description, and status. By isolating these operations, we promote a clean and intuitive interface for managing lists.

**2. ** ***Query Service:*** The query service provides a dedicated functionality for retrieving information about existing lists. This separation allows for efficient and targeted retrieval of data without the need to mix concerns related to list manipulation.

**3. ** ***Add and Remove Services:*** The addition and removal of elements from a list are encapsulated in dedicated services. This decision ensures that the core functionalities of creating and modifying lists remain focused on metadata, while adding and removing elements are treated as distinct operations.

**4. ** ***Get Service:*** The get service is designed to fetch the elements associated with a specific list. By returning a well-formatted array of cache codes, we ensure that the response is concise and aligns with the purpose of fetching list elements.

**Benefits:**

- **Modularity:** Each service encapsulates a specific set of functionalities, promoting modularity and making it easier to understand, extend, and maintain the codebase.

- **Clarity:** The separation of concerns enhances code clarity by clearly defining the responsibilities of each service. This clarity aids in debugging, code reviews, and collaboration among developers.

- **Scalability:** The modular design allows for easier scalability. New features or modifications can be seamlessly added to the system without disrupting existing functionalities.

- **Ease of Testing:** Unit testing becomes more straightforward as each service can be tested independently, ensuring that changes in one area do not inadvertently affect others.

In conclusion, this design pattern aims to strike a balance between simplicity and flexibility, facilitating efficient list management within the webservice while maintaining a clear and scalable code structure. The distinct services cater to specific functionalities, providing a foundation for future enhancements and modifications.


kojoty:
@hxdimpf ok I assume this is fully tested with code current code and db - is this right? 
Another problem is that this code is OcDe code specific. Other nodes are not ready to handle such API calls so we should add some mechanism to make it safe for other nodes and return some well formatted message that given node dies not support given API call.

I think it should also be documented in XML specification.

okapi/services/lists/add_caches/WebService.php
```diff
@@ -0,0 +1,77 @@
...
+        return array(
+            'min_auth_level' => 3
+        );
+    }
+
+    public static function call(OkapiRequest $request)
+    {
+        if (Settings::get('OC_BRANCH') != 'oc.de')
+            throw new BadRequest('This method is not supported in this OKAPI installation.');
+
```
hxdimpf, line 25: @kojoty this is the guard for functions that are platform specific

hxdimpf:
> @hxdimpf ok I assume this is fully tested with code current code and db - is this right? Another problem is that this code is OcDe code specific. Other nodes are not ready to handle such API calls so we should add some mechanism to make it safe for other nodes and return some well formatted message that given node dies not support given API call.
> 
> I think it should also be documented in XML specification.

1. thanks for reviewing.
2. yes I have fully tested the lists API on OCDE
3. There is a concept of having API methods that are platform specific. This concept includes guarding requests from not supported platforms, I have implemented this guard as per guidelines. The concept also includes a well defined markup in the documentation, via an `infotag` like this:
`<xml infotags='ocde-specific'>`
I have updated the XML for all 7 methods as  those were indeed missing, thx for pointing it out.

hxdimpf:
Addressed @kojoty's feedback and one additional issue found during review (da51761):

**Guard consistency fix:**
`create`, `delete`, `query`, and `update` were silently returning `{'success': false}` with HTTP 200 on non-OCDE installations — a misleading response that gives the caller no indication the method is unsupported. Changed all four to `throw new BadRequest(...)`, matching the pattern already used by `add_caches`, `get_caches`, and `remove_caches`. All 7 services now behave consistently.

**Docs:**
Documented the `has_lists` field in `services/apisrv/installation` (it was already returned by the code but absent from the API reference). Clients should check this field before calling any `services/lists/*` method.

The `ocde-specific` infotag and OCDE-only guards were already in place on all 7 methods from the previous commit.

hxdimpf:
Fixed the infotags format in all seven list service docs in a7cb0f0: changed `<xml infotags='ocde-specific'>` to `<xml>` with `<infotags>ocde-specific</infotags>` as a child element. The same issue was caught during review of PR #630 — the `apiref/method` service reads `$docs->infotags` as a child element, so the attribute form was silently ignored and the methods were not showing up as OCDE-specific in the API reference.

Issue #633, opened 2024-11-21
Handle invalid UUIDs in found_by parameter for OKAPI services


stefopl:
There are a lot of errors of this type appearing:

```

I suspect that all of them are coming from c:geo.

When getUserUUID() returns null, c:geo makes a request to /okapi/services/caches/shortcuts/search_and_retrieve with the found_by parameter set to unknown-user, which is not a valid UUID and causes an error.

They probably shouldn't be making the search_and_retrieve request at all when there is no valid UUID, because they won’t receive any results anyway.

I suggest implementing proper handling for cases where the UUID is invalid.

Below is a table showing the expected behavior in various cases and the current behavior:

For consistency in the examples, I used /okapi/services/caches/search/all instead of /okapi/services/caches/shortcuts/search_and_retrieve.

stefopl:
#634

PR #634, opened 2024-11-21
fix(#633): handle invalid UUIDs in found_by parameter

stefopl:


okapi/services/caches/search/SearchAssistant.php
```diff
@@ -423,10 +423,29 @@ public function prepare_common_search_params()
             try {
                 $users = OkapiServiceRunner::call("services/users/users", new OkapiInternalRequest(
                     $this->request->consumer, null, array('user_uuids' => $tmp, 'fields' => 'internal_id')));
+	            if (empty($users) || !is_array($users)) {
```
wrygiel, line 426: This should be outside (below) the surrounding try..catch.

okapi/services/caches/search/SearchAssistant.php
```diff
@@ -423,10 +423,29 @@ public function prepare_common_search_params()
...
+		        }
+	        }
+
+	        if (!empty($invalid_uuids)) {
+		        throw new InvalidParam(
+			        'found_by',
+			        "The following UUID(s) are invalid or not found: " . implode(", ", $invalid_uuids)
+		        );
+	        }
+
```
wrygiel, line 448: Indent with spaces (match the existing indentation).

wrygiel (approved):
Thanks for the fix! This looks good to me (apart from some details mentioned in the review). However, I no longer actively maintain OKAPI, so I suggest you wait at least one day before merging (perhaps there are other, more active reviewers who'd want to chip in).

stefopl:
It would be best if the fix waited for merging at least until Monday.
And if someone without a development environment would like to check, they can go to ocpl .stefo .pl/ okapi/

PR #636, opened 2026-05-08
fix(#633): handle invalid UUIDs in not_found_by parameter

stefopl:
- fix(#633): handle invalid UUIDs in not_found_by parameter
- refactor: extract duplicate found_by and not_found_by filter logic into a reusable method

### DIFF
--- a/okapi/services/caches/search/SearchAssistant.php
+++ b/okapi/services/caches/search/SearchAssistant.php
@@ -467,6 +467,23 @@ class SearchAssistant
                 throw new InvalidParam('not_found_by', $e->whats_wrong_about_it);
             }
 
+	        if (empty($users) || !is_array($users)) {
+		        throw new InvalidParam('not_found_by', "No valid users found for the given parameter.");
+	        }
+
+	        $user_uuids = explode("|", $tmp);
+
+	        $invalid_uuids = [];
+	        foreach ($user_uuids as $uuid) {
+		        if (!isset($users[$uuid])) {
+			        $invalid_uuids[] = $uuid;
+		        }
+	        }
+
+	        if (!empty($invalid_uuids)) {
+		        throw new InvalidParam('not_found_by', "The following UUID(s) are invalid or not found: " . implode(", ", $invalid_uuids));
+	        }
+
             $internal_user_ids = array_map(function ($user) { return $user["internal_id"]; }, $users);
             if (Settings::get('USE_SQL_SUBQUERIES')) {
                 $found_cache_subquery = self::get_found_cache_ids_subquery($internal_user_ids);


okapi/services/caches/search/SearchAssistant.php
```diff
@@ -815,6 +764,49 @@ public function prepare_common_search_params()
         }
     }
 
+	private function apply_found_by_filter(&$where_conds, $is_exclude = false)
```
mlekorlz, line 767: `recommended_by`, ten lines further down, has the same defect this PR fixes and is not covered.
Same on both branches:

    /okapi/services/caches/search/all?recommended_by=00000000-0000-0000-0000-000000000000
    OCPL master / this PR:  500 text/plain
    OCDE master / this PR:  500 text/plain

`SearchAssistant.php:497-507` on master is the same shape as the code being extracted here: same
`services/users/users` call, same unguarded `array_map` over the result.

It also throws with the wrong parameter name:

```php
// SearchAssistant.php:503, inside the recommended_by handler
throw new InvalidParam('not_found_by', $e->whats_wrong_about_it);
```

Folding `recommended_by` into `apply_found_by_filter()` is not a straight fit, since it needs
`get_recommended_by_sql()` rather than the `cache_id in (...)` form, so it may belong in a separate
PR. The wrong parameter name at :503 is a one-line fix either way.

mlekorlz (commented):
Tested on both branches:

- OCPL: local install from the devel dump, 56k caches, 2.7M log entries, PHP 8.1.
- OCDE: local `oc-server3` (`development`) on ddev, PHP 8.2, OKAPI pulled by composer at the same
  `3aab5717` this branch is based on.

Reproduced #633 for `not_found_by` on master and confirmed it is fixed here. Same requests against
`services/caches/search/all`:

| parameter | OCPL master | OCPL this PR | OCDE master | OCDE this PR |
|---|---|---|---|---|
| `found_by=<existing uuid>` | 200 json | 200 json | 200 json | 200 json |
| `found_by=unknown-user` | 400 json | 400 json | 400 json | 400 json |
| `found_by=00000000-0000-0000-0000-000000000000` | 400 json | 400 json | 400 json | 400 json |
| `not_found_by=<existing uuid>` | 200 json | 200 json | 200 json | 200 json |
| `not_found_by=unknown-user` | **500 text/plain** | 400 json | **500 text/plain** | 400 json |
| `not_found_by=00000000-0000-0000-0000-000000000000` | **500 text/plain** | 400 json | **500 text/plain** | 400 json |

The `search_and_retrieve` path from the issue, the one c:geo actually uses, behaves the same way on
both installs: 500 text/plain on master, 400 json here.

Both `USE_SQL_SUBQUERIES` paths are covered. OCPL sets it to true; OCDE does not set it at all, so
it takes the `Settings.php:237` default of false and runs `get_found_cache_ids()` instead of the
subquery. I also flipped it to false on OCPL to be sure. Same results everywhere.

No behaviour change on valid input. Responses are byte-identical between master and this branch on
both installs, for `found_by`, `not_found_by`, both parameters combined, and with a large `limit`.

One remark below.

